PR #8, opened 2018-01-22
add Dockerfile for the database microservice.

grinrag:
It uses default PostgreSQL image from docker hub https://hub.docker.com/_/postgres/


jdobes:
As a basic example to use as container for testing (note: it's PostgreSQL 10) it works but for production use I'd use CentOS/RHEL based images.

PR #10, opened 2018-01-24
update README for database service.

grinrag:


jdobes:
1. I suppose the line:
docker exec -it postgresql94 bash
should be:
docker exec -it vmaas_db bash

2. There is a typo in word "Download" in step 2

3. The create container command tunnels the 5432 port to the host and it's possible to connect to DB two ways from host OS:
psql -h localhost -U vmaas_user vmaas
psql -h <container IP> -U vmaas_user vmaas
maybe the tunelling is not necessary

grinrag:
1,2 and 3 fixed.

PR #15, opened 2018-01-30
add dummy crawler.

grinrag:
The real crawler will crawl via Red Hat CDN, download and parse repodata.
For now, I just add a predefined list of repositories.


jdobes:
This runs the command 16 times for each repository. It could process all repositories in single run by specifying -r multiple times. But I suppose we don't need to bother for such temporary script.

jdobes:
Also, there still needs to be Docker network setup to make this work because current "database" hostname env variable can't be resolved (#17).

PR #16, opened 2018-01-30
Added python script to serve HTTP api requests

grinrag:


jdobes:
I think we should write yum install/update/clean commands as single RUN to minimize number of layers docker needs to store:

RUN yum -y update && yum -y install epel-release & yum -y install python-ujson python2-falcon python-tornado python-psycopg2 && yum clean all

And in reposcan Dockerfile I see running "rm -rf /var/cache/yum/*". I'm not sure if it's equivalent to "yum clean all" (maybe both commands are good to run) but we should somehow unify it.

grinrag:
1. Agreed
2. `yum clean all` remove cache only for *enabled* repos, but there is can be some cache for disabled repos (not in my case), so `rm -rf /var/cache/yum/*` will remove everything. I will rewrite the command with "rm - rf ..."

PR #41, opened 2018-02-09
add product ID column

jdobes:
This table is not filled with data yet. Add INT column to store engineering product ID here.

Example:
eng_product_id: 69
name: "Red Hat Enterprise Linux Server"


grinrag:
LGTM.
I also added an issue to update database schema. https://github.com/RedHatInsights/vmaas/issues/42

PR #49, opened 2018-02-12
api: add package list to response in JSON format.

grinrag:


jdobes:
The response doesn't seem to be correct:

- There are single quotes which are not valid JSON.
- Result should be one big JSON, now each packages JSON is returned separately, output looks like [pkg1_json][pkg2_json][pkg3_json]

Issue #39, opened 2018-02-09
API: fix way of passing input data in POST request.


grinrag:
Now we are expecting to receive an input data as a json file attached to POST request. We also should expect to receive data as a body of POST request.

jdobes:
Isn't enough to support only receiving it as body (not as a file)?

For example curl should can load the file as body too:

`curl -d  '{"key1":"value1", "key2":"value2"}' -H "Content-Type: application/json" -X POST http://<FQDN>:8080/api/v1/json/`

`curl -d "@data.json" -H "Content-Type: application/json" -X POST http://<FQDN>:8080/api/v1/json/`

grinrag:
Maybe it's enough, but now we support input data as a file only. So anyway we need to implement the second way.

jdobes:
Merged.

PR #50, opened 2018-02-12
api: fix way of passing input data in POST request

grinrag:
this fixed issue #39


jdobes:
Now both passing as file and as body works, passing as file has higher priority.

PR #52, opened 2018-02-12
fix duplicate results caused by joined CVE table

### LINKED ISSUE #48: webapp: duplicates results are returned
For some nevra requests duplicated lines are responded. This is probably caused by error in DB select.

jdobes:
fixes #48


grinrag:
LGTM

PR #59, opened 2018-02-22
Update README.md with initial setup instructions

eherget:


jdobes:
I see you are installing docker from official docker repository, this repository contains most recent version. There is also an older docker (1.13) shipped in Fedora repository and RHEL 7 Extras, I'm using this version and probably other guys too. I'm not sure if there is any real difference for us between these two versions (probably not) but I'd stick with Fedora/Red Hat/CentOS builds in default instructions or at least mention both options here.

eherget:
I have updated the README.md to use docker from fedora repos instead of the docker-ce repo.

PR #65, opened 2018-02-28
api: issue 57, change way of parsing RPM package name

grinrag:


jdobes:
Why is the `name-epoch:ver-rel.arch` prefferred instead of current `epoch:name-ver-rel.arch`?

grinrag:
We discussed this with @MichaelMraka. For example, `rpm -qi` doesn't understand `epoch:name-ver-rel.arch` format.

MichaelMraka:
True, neither `dnf`. It's just `yum`'s invention not used anywhere else.

PR #81, opened 2018-03-05
api: updates module: fix some minor issues

grinrag:


webapp/updates.py
```diff
@@ -158,6 +158,19 @@ def process(filename, cursor):
     return [ merge_dict(x, {'name' : n}) for x in res ]
 
 
+"""
```
jdobes: The docstring should be placed in the function, not before the function - https://www.python.org/dev/peps/pep-0257/.

Issue #70, opened 2018-03-02
No longer get results from API call


eherget:
I've recently rebased my fork off of RedHatInsights/vmaas and now I no longer get results when I hit the API.  I've checked the database manually and at least 2 packages (all that I checked) submitted have later version packages in the db.  The response from the API call simply lists the same packages I sent in.

I reset my git repo to various commits and found that this commit 6be821afad80d81ed1b2ea8c70c279f506fd3a58 (api: add method to process input packages list and make result in format:...) is the first commit where I get no results.

In debugging this issue, it seems the database image is somehow being pre-populated with data as the first time reposcan runs on a freshly built image/container, it skips downloading from all 16 repos.  I'm not sure if this is related, but my concern is the pre-populated database could be missing values in newly added fields, etc.

my pkgfile.json contains:
{
    "package_list": [
        "qt-settings-19-22.4.el7.noarch",
        "perl-Pod-Checker-1.50-1.el7.noarch",
        "bash-4.2.46-20.el7_2.x86_64",
        "postgresql-9.2.10-1.el7.x86_64",
        "xorg-x11-proto-devel-7.7-13.el7.noarch"
    ]
}

The command to make the API call and view the response is:
curl -X POST -d @pkgfile.json -H "Content-Type: application/json" http://localhost:8080/api/v1/json | python -m json.tool | less

the API response from current RedHatInsights/master is:
{
    "bugzilla_list": {},
    "cve_list": {},
    "errata_list": {},
    "nevra_list": {},
    "package_list": {
        "bash-4.2.46-20.el7_2.x86_64": [],
        "perl-Pod-Checker-1.50-1.el7.noarch": [],
        "postgresql-9.2.10-1.el7.x86_64": [],
        "qt-settings-19-22.4.el7.noarch": [],
        "xorg-x11-proto-devel-7.7-13.el7.noarch": []
    },
    "repo_list": {}
}

if I reset my git repo to the commit before the one mentioned above, the API response is:
{
    "bugzilla_list": {},
    "cve_list": {},
    "errata_list": {},
    "nevra_list": {},
    "package_list": {
        "bash-4.2.46-20.el7_2.x86_64": [
            [
                "bash-4.2.46-28.el7",
                "RHSA-2017:1931",
                "http://pulp-read.dist.prod.ext.phx2.redhat.com/content/dist/rhel/server/7/7Server/x86_64/os/"
            ],
            [
                "bash-4.2.46-28.el7",
[...]

PR #82, opened 2018-03-06
api: updates: fix for the issue #70

grinrag:
Errata can be associated with several repos, so now we are checking package repo IDs matching to some errata repo IDs.


jdobes:
Merging now to fix the empty result but let's make this function more readable, there is too many nested cycles and conditions.

PR #84, opened 2018-03-06
Developer setup

MichaelMraka:
Slightly modified compose to make development easier.

How to instructions added to the end of README.

### DIFF
--- a/docker-compose.devel.yml
+++ b/docker-compose.devel.yml
@@ -1,0 +1,25 @@
+version: '3'
+
+services:
+  database:
+    volumes:
+      - ./database:/git
+    security_opt:
+      - label=disable
+    working_dir: /git
+
+  reposcan:
+    volumes:
+      - ./reposcan:/git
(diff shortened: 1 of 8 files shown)


grinrag:
Looks like the commit `808870e` is just a fixup for the first commit `ae65a27`. Could, you, please, merge them?

docker-compose.devel.yml
```diff
@@ -0,0 +1,38 @@
...
+      file: docker-compose.yml
+      service: database
+    volumes:
+      - ./database:/git
+    security_opt:
+      - label=disable
+    working_dir: /git
+
+  reposcan:
+    extends:
```
ggainey: 'extends' isn't supported by docker v3 :(

https://docs.docker.com/compose/compose-file/compose-versioning/#version-3

MichaelMraka: Oh My Docker... I've naively tested it on my v2 compose :(

PR #97, opened 2018-03-08
Fix error in JsonHandler

jiridostal:


eherget:
This fixed the issue I was having!

PR #107, opened 2018-03-12
api: updates: add filtering by reponame list.

grinrag:


webapp/updates.py
```diff
@@ -272,5 +285,9 @@ def process_list(cursor, data):
     response = {
         'update_list': answer,
     }
+
+    if provided_repo_ids is not None:
+        response.update({'repo_lsit': provided_repo_names})
```
MichaelMraka: typo "lsit" -> "list"
moreover we should we should agree together to use either repo or repository all across code consistently

PR #111, opened 2018-03-13
api: updates: for GET request handling we need to match packages name.

grinrag:
additional changes for the commit 3eb717cc1b7cc3c01c8af58a86d5e6b93d4cfea7

### DIFF
--- a/webapp/app.py
+++ b/webapp/app.py
@@ -76,7 +76,8 @@ class Application(tornado.web.Application):
     def __init__(self):
         handlers = [
             (r"/?", DocHandler),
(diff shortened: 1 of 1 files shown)


webapp/app.py
```diff
@@ -76,7 +76,7 @@ class Application(tornado.web.Application):
     def __init__(self):
         handlers = [
             (r"/?", DocHandler),
-            (r"/api/v1/updates/?", UpdatesHandler),
+            (r"/api/v1/updates/?.*", UpdatesHandler),
```
MichaelMraka: should be without '?' otherwise it matches e.g. /updateswhatever

PR #112, opened 2018-03-13
api: url for GET request should be without '/' at the end.

grinrag:
We are using rfind('/') method of a string and then cut cve/package/repo name, so URL for a GET request should be without '/' at the end.


ggainey:
Resolved merge-conflict and commited as 7b95b517c4bf8ca98f7c068a23b44f1a5523df88 - will try not to push buttons on github ui in future...

PR #121, opened 2018-03-14
Enable pylint in webapp

MichaelMraka:


grinrag:
First of all, I'm completely against of commits, which just fix pylint issues because these commits are rewriting the history. Then when I make `git blame` I don't see the real reason for some commits, instead of this I just see `pylint fixes` for many lines of code and I need to checkout to the previous commit to see the history. Possible solution - disable pylint warnings in such lines and add  FIXME message. 
Docstrings are very useful.  Unused imports and shebang also could be removed.

ggainey:
On my last check, it all Just Worked. accepted, with puzzlement

PR #122, opened 2018-03-14
SQL reformating & variable binding

MichaelMraka:


grinrag:
All commits with the same commit message. Could, you, please, merge them into one, since the reason of all these commits is the same.

Issue #119, opened 2018-03-13
Errata API missing data for fields


eherget:
There are four fields that are defined in the VMaaS API details draft document for which I could not find existing data in the database.  This issue is to add the fields to the database and update the Errata API implementation to include these fields in the response.

- type
- summary
- bugzilla list
- reference list

eherget:
PR #141 adds type and summary fields to errata.

eherget:
PR #157 finishes this issue.

PR #157, opened 2018-03-20
Add more errata data

eherget:
This finishes Issue #119.
It also reworks how the various lists of info in an erratum are retrieved - it makes 3 db queries to populate lists in all errata instead of 3 db queries per erratum.


eherget:
Please check my email before merging.

PR #161, opened 2018-03-22
simple test for API output

MichaelMraka:
mainly for manual sanity tests
not suitable for travis because it requires working application incl. loaded data


ggainey:
Looks good (as in, works, altho it already finds 'errors' due to data-model improvements). Ultimately, the test-harness should drive off of versioned db-snapshots, so we'd start with a known data set. But that is for a later PR :) Merging.

Issue #166, opened 2018-03-26
Wrong Content-Type in repo sync and cve sync response


mkoura:
For both ``/api/v1/sync/repo`` and ``/api/v1/sync/cve`` the ``Content-Type`` of the response is ``text/html; charset=UTF-8`` where it should be ``application/json; charset=UTF-8``.

The response data is JSON:
```json
{
	"success": true,
	"msg": "Repo sync task started."
}
```

It's important to set the Content-Type correctly so automated tools can handle the data contained in the response.

PR #171, opened 2018-03-26
166 - Teach reposcan to set Content-Type to app/json

ggainey:
There has to be a better/more-std/more-pythonic way to do this

tornado should auto-recognize write(dict) as Content-type app/json - what's the best way to make ResponseMsg return a dict instead of a string that just happens to look a lot like json?

### DIFF
--- a/reposcan/reposcan.py
+++ b/reposcan/reposcan.py
@@ -107,14 +107,15 @@ class SyncTask:
         return cls._running
 
 
-class ResponseMsg: # pylint: disable=too-few-public-methods
-    """Object used as API response to user, represented as JSON string."""
+class ResponseJson(dict): # pylint: disable=too-few-public-methods
+    """Object used as API response to user, represented as JSON"""
     def __init__(self, msg, success=True):
-        self.msg = msg
-        self.success = success
+        self['msg'] = msg
+        self['success'] = success
 
     def __repr__(self):
-        return json.dumps({"success": self.success, "msg": self.msg})
+        return json.dumps({"success1": self.success, "msg2": self.msg})
+
 
 
 class SyncHandler(RequestHandler):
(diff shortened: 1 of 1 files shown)


MichaelMraka:
What about
```
class ResponseJson(dict):
    """Object used as API response to user, represented as JSON."""
    def __init__(self, msg, success=True):
        self["msg"] = msg
        self["success"] = success
```
and then instead of 
```
self.write(msg)
```
or
```
self.write(str(ResponseMsg(msg, success=False)))
```

use
```
self.write(ResponseJson(msg, success=False)))
```

jdobes:
I agree, Tornado should automatically set JSON Content-Type when dictionary is passed to write method (we do this in webapp already). Currently there is string passed - `str(ResponseMsg(...))` with explicit JSON conversion in `__repr__` method.

ggainey:
Have to fix some pylint issues, do not merge

reposcan/reposcan.py
```diff
@@ -110,11 +110,12 @@ def is_running(cls):
 class ResponseJson(dict): # pylint: disable=too-few-public-methods
     """Object used as API response to user, represented as JSON"""
     def __init__(self, msg, success=True):
+        super(ResponseJson, self).__init__()
         self['msg'] = msg
         self['success'] = success
 
     def __repr__(self):
-        return json.dumps({"success1": self.success, "msg2": self.msg})
+        return json.dumps({"success1": self['success'], "msg2": self['msg']})
```
MichaelMraka: Isn't just ```return json.dumps(self)``` better/easier?

jdobes: Or we can remove the `__repr__` method completely, we don't call str() anymore.

ggainey: ugh, I'm an idiot - mim, yeah, should just be calling dumps on self. jdobes, I'll leave __repr__ in there, we're not using str(ResponseJson) now, but may want//need to.

MichaelMraka:
squashed into dfd83f475d5ad84ebdb2cd4617131b2f6a471de7
fixes #166

Issue #145, opened 2018-03-16
Disallow empty strings in database


MichaelMraka:
Any text/varchar column in  db should not allow empty strings.
Add constraints to disallow it.

PR #174, opened 2018-03-27
145 - Disallow empty strings in db

ggainey:
* Checkconstraint for empty on all TEXT fields
 * Teach cve_store to store null instead of '' when no-url-found
 * Teach updateinfo to store null if field is empty/white-space-only
 * Teach repository_store to store null instead of empty-string for key

### DIFF
--- a/database/vmaas_db_postgresql.sql
+++ b/database/vmaas_db_postgresql.sql
@@ -29,7 +29,7 @@ create or replace function isdigit(ch CHAR)
 $$ language 'plpgsql';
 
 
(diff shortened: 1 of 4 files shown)


ggainey:
Note that some of the data we grab is imperfect - e.g., there are Fedora updates that have no description-content. We prob need to go over the db and make sure that NOT NULL is only used in places where things really and truly MUST BE PRESENT (e.g., names/labels)

MichaelMraka:
merged as 7f0d744681c34a008d10db3d45c641275c976b68
fixes #145

PR #179, opened 2018-03-28
Repo and basearch filtering for /updates

MichaelMraka:
* Changes to make repo and basearch filtering on updates.
* updates.py code improvements

PR #180, opened 2018-03-28
find updates in repositories of same product

jdobes:
before:
- only updates from same repository are found

after:
- updates from repositories provided by same product are found (allows e.g. find update for pkg from optional in base repo and vice versa)
- releasever of original repo and update repo have to be same (prevents e.g. to update el6 package to el7 package because el6 and el7 are same product), it's possible to enhance it in future - specify allowed releasevers in request similarly as we specify repo labels


jdobes:
Let's check #179 first and rebase this PR then. There are going to be conflicts.

Issue #120, opened 2018-03-14
Fix pylint issues in webapp


MichaelMraka:
Currently I've (temporarily!) disabled pylint warnings in webapp:
************* Module errata
R: 10, 4: Too many arguments (9/5) (too-many-arguments)
R: 97, 4: Too many local variables (18/15) (too-many-locals)
************* Module updates
R: 73, 4: Too many local variables (57/15) (too-many-locals)
R:271, 8: Too many nested blocks (8/5) (too-many-nested-blocks)
R: 73, 4: Too many branches (49/12) (too-many-branches)
R: 73, 4: Too many statements (134/50) (too-many-statements)

Needs to be resolved and remove '#pylint disable' comments.

eherget:
This issue was to resolve pylint issues in webapp.  Specifically, to make modifications so that "# pylint disable:..." comments were no longer necessary.

I had done some work on this previously (on the errata api), and then put it on hold as we prepared for the Insights demo at the Red Hat conference.  Now that the demo is over, I'm questioning if there is still a need for this work.  We have several "#pylint disable:..." comments throughout the code for things like too-many-arguments, too-many-locals, too-many branches, etc. Refactoring the code to get rid of the need for *all* of these pylint disable comments seems like a lot of work, for not much gain, and might result in some unnecessary things like adding extra class methods just for the sake of dropping a pylint warning.

My suggestion would be to close this issue as overly broad.  Going forward, we would open issues for specific pylint warnings we believe indicates an actual issue we should address.

MichaelMraka:
I agree.
I'd also globally disable some more standard pylint checks we don't enforce (e.g. too-few-public-methods - disabled 12 times, too-many-locals - 6 times, too-many-arguments 5 times)...

ggainey:
General issue addressed in #283

PR #188, opened 2018-04-02
Refactor to prevent pylint warnings in errata api

eherget:
This PR addresses Issue #120 for the errata api - it gets rid of the need to disable pylint warnings for too-many-arguments and too-many-local-variables.

This also changes errata api to use RealDictCursor from the psychopg2.extras package.  This allows us to use the dictionaries returned directly from the sql queries in the api response.  Take a look and see if we want to apply this across all APIs as it may be a slight performance increase in addition to removing the need to disable pylint warnings for too-many-arguments and too-many-local-variables.


eherget:
Closing this pull request.  Submitted new PR #189 that can be rebased and merged.

Issue #23, opened 2018-02-01
reposcan: support configuration of values


jdobes:
There are (and will be) some hard coded values in code like number of threads to download etc. that are good to be configurable using some config file or some container-friendly way.

jdobes:
In docker are commonly used env variables for configuration.

ggainey:
Candidates so far:

Definitely should be configurable:
downloader.py THREADS

Maybe configurable?
downloader/unpacker.py CHUNK_SIZE (at minimum, get this from one place instead of two)
batch_list.py DEFAULT_BATCH_SIZE
cve_controller.py YEAR_SINCE

Probably not useful:
downloader.py VALID_HTTP_CODES 
package_store.py CHECKSUM_TYPE_ALIASES

jdobes - thoughts?

jdobes:
THREADS, CHUNK_SIZE, DEFAULT_BATCH_SIZE and YEAR_SINCE makes sense to make configurable using `vmaas/conf/reposcan.env`

PR #191, opened 2018-04-03
docker-compose/OpenShift fixes

jdobes:
- rebuild Swagger UI image to work in OpenShift
- split single docker-compose.env into multiple files in conf/ directory, allow to have application-specific configuration options (can be used in #23), these env files are translated into separate OpenShift Config Maps


ggainey:
Testing this on latest-master-from-docker, I often see
```
Code | Details
-- | --
Undocumented | TypeError: Failed to fetch
```
in response to a request that succeeded. Specifically, from http://localhost:8000/?urls.primaryName=reposcan#/default/get_sync_cve , testing "GET /sync/cve" or just "GET /sync" I see it, even though logs show everything is fine, and using curl directly returns a 200.

Not sure how serious this is, and I don't see how the changes here could be causing it - thoughts?

jdobes:
It's because of missing "Access-Control-Allow-Origin" header in response, Swagger UI requires it, it's actually already set for the /api/v1/apispec and here is the response displayed. It's not related to these changes, I created #192 .

PR #196, opened 2018-04-04
Added modified_since filter to /cves and /errata

MichaelMraka:
Also unified timestamp parsing across the apllication.


jdobes:
Since we have apidoc now, please add the `modified_since` to the example input and output definitions.

PR #213, opened 2018-04-12
Fix #23 - make some hardcoded values tuneable via env

### LINKED ISSUE #23: reposcan: support configuration of values

ggainey:
#23


jdobes:
in `conf/reposcan.env` and `reposcan/Dockerfile` is `DEFAULT_BATCH_SIZE` defined but in `reposcan/common/batch_list.py` `os.getenv('BATCH_SIZE'...` is called.

ggainey:
That...is a good catch. Update incoming.

PR #224, opened 2018-04-17
Fix #207 api: cves: add ability to get all CVEs by regular expression.

### LINKED ISSUE #207: Add ability to get all cve/errata modified-since
As discussed 10-APR - would like to be able to ask for all errata/cves modified-since a given date

Notes:
 - consider data-size - pagination? just accept that some responses are going to be huge?
 - specify date-range

grinrag:


grinrag:
Thank you, Grant!
1) I updated docs and fixed GET handler. 
2) The patch adds regular expression not a globex, so http://localhost:8080/api/v1/cves/CVE-2013-7* - returns all CVE-2013, because `*` means zero or more repeats of the previous character. This is correct.

PR #229, opened 2018-04-18
Fix #223 - Add limited users to different components

### LINKED ISSUE #223: Basic database-user sanitation
We're going to expose this on Teh Internets. Basic security sanity requires least-privilige access. So we're going to need three vmaas-users with different access-levels:

* vmaas_reader : used by webapp, has RO-access to all tables
* vmaas_writer : used by reposcan and plugins, has INSERT/UPDATE/DELETE access to all tables
* vmaas_admin : used *only* for database init, has dba access (specifically, create users/tables/roles)

ggainey:
- vmaas_admin is used by database to create vmaas DDL
- vmaas_writer is used by reposcan to insert/update/delete DML
- vmaas_reader has select-only access to tables, used by webapp

Notes:
* Currently we create everything into 'public' schema; may want to have
a specific 'vmaas' schema.

* CREATE USER has to run as the postgres user initially; hence the
database/vmaas_user_create_postgresql.sql being run in database/init_schema.sh


jdobes:
I think this doesn't work as intended because of docker-compose ENV variables priorities overriding.
Highest priority have variables is env file, then variables in Dockerfile, then defaults in code.

Example:

webapp/database.py:
DEFAULT_USER = "vmaas_reader"
DEFAULT_PASSWORD = "vmaas_reader_pwd"

webapp/Dockerfile:
ENV POSTGRESQL_USER=vmaas_reader
ENV POSTGRESQL_PASSWORD=vmaas_reader_pwd

conf/database-connection.env:
POSTGRESQL_USER=vmaas_admin
POSTGRESQL_PASSWORD=vmaas_admin_passwd

In this case webapp will use admin user instead of reader user.

MichaelMraka:
You can create conf/database-connection-reader.env with reader user and add it into docker-compose.yml. It will overwrite default db user (last env file wins).

jdobes:
There are still default credentials in `reposcan/reposcan.py`, should be changed to vmaas_writer

reposcan/reposcan.py
```diff
@@ -37,7 +37,7 @@
 def init_db():
     """Setup DB connection parameters"""
     DatabaseHandler.db_name = os.getenv('POSTGRESQL_DATABASE', "vmaas")
-    DatabaseHandler.db_user = os.getenv('POSTGRESQL_USER', "vmaas_user")
+    DatabaseHandler.db_user = os.getenv('POSTGRESQL_USER', "vmaas_writer")
     DatabaseHandler.db_pass = os.getenv('POSTGRESQL_PASSWORD', "vmaas_passwd")
```
jdobes: Please change the default password for the vmaas_writer.

ggainey: Having code default the user if it isn't set in the env is one thing - but the passwd shouldn't be hard-coded. That's why I left that as I did.

jdobes: It's only a default password (overridden by Dockerfile value and overridden by ENV map file value), in `webapp/database.py` you are changing the default pw too.

PR #231, opened 2018-04-20
api: cves: adds a temporary workaround for the CVEs response pagination.

grinrag:
This is a possible workaround for CVEs pagination in response. Docs are not updated since we are not going to merge this PR right now.


ggainey:
I have lost track - are we going to merge this so it hits stable, or not?

ggainey:
Closing - obsoleted by #290

PR #244, opened 2018-04-25
Add package_name table, missing indexes and improve performance of reposcan and webapp

jdobes:


ggainey:
Pre-PR, 600-RPM RHEL6 system against DB w/17 repos, 1.8 secs. Post-PR, 0.758 sec. 60% is a worthwhile perf gain

PR #246, opened 2018-04-26
fix #190 - add ansible tasks to do more things in OpenShift automatically

### LINKED ISSUE #190: Automate OpenShift routes management
OpenShift is using [routes](https://docs.openshift.com/enterprise/3.0/architecture/core_concepts/routes.html) to expose services to outside. This routes are created/removed/managed manually now. For Swagger UI it's also needed to update API_URLS config to point to these routed URLs to make it work in OpenShift (default localhost:port works only in docker compose).

jdobes:


ggainey:
Looks good and didn't break my dev-env - I will 'assume' it won't break OS CI env :)

PR #261, opened 2018-05-15
Make updates api db query more efficient

eherget:
This changes the updates api to do a single db query where it used to do one db query per package specified in the request.  This implementation does the evr comparison in the python/webapp code.  The rpmver_array() and other evr related functions and types are no longer required in the database.

In my tests, what used to take 22 seconds now takes 15 with this change.

### DIFF
--- a/webapp/updates.py
+++ b/webapp/updates.py
@@ -22,6 +22,90 @@ JSON_SCHEMA = {
     }
 }
 
+class RpmVerArray(object):
+    """ Class for processing rpm version and release values for
+        comparison to other rpm version and release values. """
+    def __init__(self, value):
+        self.value = value
+        self.remaining_value = value
+        self.ver_array = []
+        self.current_array_index = 0
+
+    def _parse_next_tuple(self):
+        while self.remaining_value and not self.remaining_value[0].isalnum():
+            self.remaining_value = self.remaining_value[1:]
+        if not self.remaining_value:
+            return None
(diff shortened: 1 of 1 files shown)


eherget:
I need to add that I did not remove the mentioned evr functions and types defined in the database.  Should we accept this change, that can be a follow up issue/pr

ggainey:
So was doing some before/after comparisons with this code, and I noticed one thing - pre-change, NEVRA rpcbind-**0.2.0-13.el6_9.1**.x86_64 finds no available-updates; post-change, it finds rpcbind-**0.2.0-13.el6_9**.x86_64 . Clearly this is an edge case, but I'm not sure which one is correct - Eric, could you take a look at this?

eherget:
I will take a look at this rpcbind example issue

eherget:
Good catch, ggainey.  My logic was incorrect when comparing a value from one evr to a null from the other (9.1 v 9, for example, the '1' on the left is greater than the non-existent corresponding value on the right)

This is now fixed and this PR can be accepted once ginrag completes his review.

grinrag:
@eherget I tested the changes. So far I see a difference in the answers produced by webapp based on `RedHatInsights:master`  and webapp based on `eherget:efficient-updates-db-queries`.

I tested on this input:
[pack.txt](https://github.com/RedHatInsights/vmaas/files/2013152/pack.txt)

I see performance improve:
`0.647s` against `1.150s`

Issue #263, opened 2018-05-22
webapp gets into corrupted state


jdobes:
found in logs:

```ERROR:tornado.access:500 POST /api/v1/updates/ (10.130.10.1) 16181.89ms
ERROR:tornado.application:Uncaught exception POST /api/v1/updates/ (10.130.4.1)
HTTPServerRequest(protocol='http', host='webapp-vmaas-ci.1b13.insights.openshiftapps.com', method='POST', uri='/api/v1/updates/', version='HTTP/1.1', remote_ip='10.130.4.1', headers={'Content-Length': '45221', 'Forwarded': 'for=185.108.120.60;host=webapp-vmaas-ci.1b13.insights.openshiftapps.com;proto=http', 'X-Forwarded-Host': 'webapp-vmaas-ci.1b13.insights.openshiftapps.com', 'X-Forwarded-Port': '80', 'X-Forwarded-For': '185.108.120.60', 'User-Agent': 'tsung', 'Host': 'webapp-vmaas-ci.1b13.insights.openshiftapps.com', 'X-Forwarded-Proto': 'http', 'Content-Type': 'application/x-www-form-urlencoded'})
Traceback (most recent call last):
  File "/usr/lib64/python2.7/site-packages/tornado/web.py", line 1368, in _stack_context_handle_exception
    raise_exc_info((type, value, traceback))
  File "/usr/lib64/python2.7/site-packages/tornado/stack_context.py", line 313, in wrapped
    ret = fn(*args, **kwargs)
  File "/usr/lib64/python2.7/site-packages/tornado/web.py", line 1580, in future_complete
    f.result()
  File "/usr/lib64/python2.7/site-packages/tornado/concurrent.py", line 214, in result
    raise_exc_info(self._exc_info)
  File "/usr/lib64/python2.7/site-packages/tornado/gen.py", line 876, in run
    yielded = self.gen.throw(*exc_info)
  File "/app/app.py", line 234, in post
    res = yield self.background_process(endpoint='/updates', data=data)
  File "/usr/lib64/python2.7/site-packages/tornado/gen.py", line 870, in run
    value = future.result()
  File "/usr/lib/python2.7/site-packages/concurrent/futures/_base.py", line 398, in result
    return self.__get_result()
  File "/usr/lib/python2.7/site-packages/concurrent/futures/thread.py", line 55, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/app/app.py", line 55, in background_process
[...]

PR #277, opened 2018-05-28
#263 close every DB connection since we don't have pool yet.

genalt:


eherget:
Prior to this change, the app appears to close connections after each request.  I test this by running several queries against the api's that use the db (cve, errata, etc) and then do `netstat -anp | grep 5432` and usually see only the db listening on port 5432.

Is there a specific use case for this change that I can use to confirm it fixes something?

eherget:
I see the details of the problem reported in issue #263.  I can't make it fail like that when running 200 api request and I assume that is because I can't perform 100+ concurrent request.  However, I see no issues running against this new code, so I will go ahead and merge.

PR #283, opened 2018-05-30
Issue 120 - disable a few pylint checks in pylintrc

eherget:
… to clean up disable comments in code.

NOTE: There are a lot of to disable "arguments-differ".  Do we want to disable this in pylintrc as well?  Shouldn't we be able to make the args the same as the class from which we have extended?


MichaelMraka:
> NOTE: There are a lot of to disable "arguments-differ". Do we want to disable this in pylintrc as well? Shouldn't we be able to make the args the same as the class from which we have extended?

I find this one useful so I'm against disabling it globally.
I've tried to fix the issues (basically  by ```def get(self):``` -> ```def get(self, *args, **kwargs):```) but webapp stopped working :(. Maybe we can try it again after @grinrag's changes in tornado app.

PR #288, opened 2018-05-31
webapp: run several tornado.httpserver.HTTPServer instances.

grinrag:
This give us some performance boost.

### DIFF
--- a/conf/webapp.env
+++ b/conf/webapp.env
@@ -1,1 +1,2 @@
 REPOSCAN_WEBSOCKET_URL=ws://reposcan:8081/notifications
+MAX_VMAAS_SERVERS=2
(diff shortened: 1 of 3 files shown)


jdobes:
1. there is a bug - app doesn't connect to websocket because nothing after `IOLoop.instance().start()` command is executed
2. please make MAX_WORKERS and MAX_SERVERS configurable - set default env variables in Dockerfile and env file

Otherwise it works as expected - spawns 2 app.py processes, each consuming 700-800 MB of memory. It can be useful for scaling in local docker compose environment, not much useful in OpenShift.

grinrag:
1. Moved connection before start instance of IOLoop. Now see in log-file:
`vmaas-webapp | Connected to: ws://reposcan:8081/notifications`, number of such messages equals to the number of Servers.

2. Add environment variable 
`ENV MAX_VMAAS_SERVERS=2`
`MAX_WORKERS` is equal to the number of threads, it became pointless, so removed them at all.

PR #294, opened 2018-06-06
Make errata api use Phase II design

eherget:
I have cut-n-pasted the join_packagename() method from webapp/utils.py to reposcan/common/package_utils.py.  I'll open an issue that this should possibly be move to a utils or lib that gets used by both webapp and reposcan.

### DIFF
--- a/reposcan/exporter.py
+++ b/reposcan/exporter.py
@@ -195,6 +195,50 @@ class DataDump:
                                                 set()).add(repo_id)
                 dump.update(errataid2repoids)
 
+        # Select errata detail for errata API
+        errataid2cves = {}
+        with self._named_cursor() as cursor:
+            cursor.execute("""SELECT errata_cve.errata_id, cve.name
+                                FROM cve
+                                JOIN errata_cve ON cve_id = cve.id
+                           """)
+            for errata_id, cve_name in cursor:
+                errataid2cves.setdefault(errata_id, []).append(cve_name)
+        errataid2pkgid = {}
+        with self._named_cursor() as cursor:
+            cursor.execute("""SELECT errata_id, pkg_id FROM pkg_errata""")
+            for errata_id, pkg_id in cursor:
+                errataid2pkgid.setdefault(errata_id, []).append(pkg_id)
+        errataid2bzs = {}
+        errataid2refs = {}
+        with self._named_cursor() as cursor:
+            cursor.execute("SELECT errata_id, type, name FROM errata_refs")
+            for errata_id, ref_type, ref_name in cursor:
+                if ref_type == 'bugzilla':
+                    errataid2bzs.setdefault(errata_id, []).append(ref_name)
+                else:
+                    errataid2refs.setdefault(errata_id, []).append(ref_name)
+        # Now pull all the data together for the dump
+        with self._named_cursor() as cursor:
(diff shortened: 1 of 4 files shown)


reposcan/exporter.py
```diff
@@ -194,6 +195,62 @@ def dump_errata(self, dump):
...
+                url = "https://access.redhat.com/errata/%s" % e_name
+                dump["errata_detail:%s" % e_name] = (synopsis, summary, e_type,
+                                                     e_severity, description,
+                                                     solution, issued, updated,
+                                                     errataid2cves.get(errata_id, []),
+                                                     errataid2pkgs.get(errata_id, []),
+                                                     errataid2bzs.get(errata_id, []),
+                                                     errataid2refs.get(errata_id, []),
+                                                     url)
+
```
jdobes: I think this service should not export details about every errata found in DB into cache but only security. All selects should have something like `where errata_id in %s` (see other selects) to filter these relevant errata.

reposcan/exporter.py
```diff
@@ -194,6 +195,62 @@ def dump_errata(self, dump):
...
+                                     evr.release,
+                                     arch.name
+                                FROM pkg_errata
+                                JOIN package ON package.id = pkg_errata.pkg_id
+                                JOIN package_name ON package_name.id = package.name_id
+                                JOIN evr ON evr.id = package.evr_id
+                                JOIN arch ON arch.id = package.arch_id
+                           """)
+            for errata_id, p_name, p_epoch, p_ver, p_rel, a_name in cursor:
+                packagename = join_packagename(p_name, p_epoch, p_ver, p_rel, a_name)
```
jdobes: I'd export only errataid to package ids mapping in the dump and then create full package name in webapp using other cached data to make database dump and memory footprint smaller.

ggainey: Altho, the whole point to the export process is to increase the webapp's speed at the expense of memory. In that context, it makes more sense for the export to have as much precomputed as possible.

eherget:
Issue #299 likely applies to this PR as well.  Hold off on merging until the issue is resolved and I'll update this PR to fix the same issue.

reposcan/exporter.py
eherget: I had the same thought as @ggainey, that pre-computing the package name made sense here for performance reasons.  In other places (updates API) it likely makes sense to keep the package data separate for query filtering/search purposes.

reposcan/exporter.py
eherget: I included all errata in the dump because I thought we had discussed previously having a flag that would return all errata, not just security ones.

However, I do see that the default behavior is to only return security errata and I did not implement that filter in the default case as I should have.  I'll update the PR to do that.

reposcan/exporter.py
jdobes: Well, it depens if there is any noticeable performance hit caused by building the NEVRA on webapp side.

eherget:
I have updated this commit to use join_packagename on the webapp side.  The performance hit is minimal and avoids duplicating code in two places.

Issue #299 is still being worked on.  It's fix for date comparison will need to be applied to any api converted to Phase II approach that supports modified_since.  As such, I don't think we need to wait for that fix to merge this PR.

This should be ready for merging.

Issue #306, opened 2018-06-12
Hot cache has not unique keys across the forks


psegedy:
In one fork you have cached response for `bash-0:4.2.46-20.el7_2.x86_64` and in the other response for the same package, but with repository_list/releasever/basearch filter. Next response for this package depends on from which fork will you get cached answer.

Response without filters
```zsh
(.vmaas_venv) ➜  vmaas_tests git:(master) ✗ curl http://localhost:8080/api/v1/updates/bash-0:4.2.46-20.el7_2.x86_64 | python -m json.tool
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   719    0   719    0     0  59916      0 --:--:-- --:--:-- --:--:-- 59916
{
    "update_list": {
        "bash-0:4.2.46-20.el7_2.x86_64": {
            "summary": "The GNU Bourne Again shell",
            "description": "The GNU Bourne Again shell (Bash) is a shell or command language\ninterpreter that is compatible with the Bourne shell (sh). Bash\nincorporates useful features from the Korn shell (ksh) and the C shell\n(csh). Most sh scripts can be run by bash without modification.",
            "available_updates": [
                {
                    "package": "bash-4.2.46-28.el7.x86_64",
                    "erratum": "RHSA-2017:1931",
                    "repository": "rhel-7-server-rpms",
                    "basearch": "x86_64",
                    "releasever": "7Server"
                },
                {
                    "package": "bash-4.2.46-28.el7.x86_64",
                    "erratum": "RHSA-2017:1931",
                    "repository": "rhel-7-workstation-rpms",
                    "basearch": "x86_64",
                    "releasever": "7Workstation"
                }
            ]
        }
    }
}

```

Response with filters
```zsh
(.vmaas_venv) ➜  vmaas_tests git:(master) ✗ cat pkg_filters.json 
{
  "package_list": ["bash-0:4.2.46-20.el7_2.x86_64"],
  "repository_list": ["rhel-7-server-rpms"],
  "releasever": "7Server",
  "basearch": "i386"
}
(.vmaas_venv) ➜  vmaas_tests git:(master) ✗ curl -d @pkg_filters.json http://localhost:8080/api/v1/updates | python -m json.tool         
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
[...]

PR #307, opened 2018-06-12
issue #306 api: updates: before put response into cache, we should ta…

grinrag:
…ke into account what repos were passed in the input JSON.

### DIFF
--- a/webapp/updates.py
+++ b/webapp/updates.py
@@ -2,6 +2,7 @@
 Module to handle /updates API calls.
 """
 
+import hashlib
 from jsonschema import validate
 
 from cache import REPO_LABEL, REPO_BASEARCH, REPO_RELEASEVER, REPO_PRODUCT_ID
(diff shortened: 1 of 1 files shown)


webapp/updates.py
```diff
@@ -347,12 +347,17 @@ def process_list(self, data):
             'update_list': {},
         }
 
+        # Get list of valid repository IDs based on input paramaters
+        available_repo_ids = self._process_repositories(data, response)
+
+        repo_ids_key = hashlib.md5('_'.join([str(id) for id in sorted(available_repo_ids)]).encode('utf-8')).hexdigest()
```
MichaelMraka: _ is a valid character in repo label so something else (space or comma?)  would be better separator.

MichaelMraka:
The very same issue apply also for releasever. We need to take it into account when building cache.

webapp/updates.py
grinrag: I'm joining repo IDs, so just numbers. Then it goes to the hash function.

grinrag:
self._process_repositories() already takes into account `'repository_list'`, `'releasever'`, `'releasever'`

webapp/updates.py
MichaelMraka: I see, then it's ok.

MichaelMraka:
Ah, I overlooked it, now it seems ok to me.

PR #332, opened 2018-06-21
add healthchecks for openshift deployment

jdobes:
1. defines HealthHandler in both webapp and reposcan used for HTTP GET pinging by probes, currenly it only returns 200 when API is available, other code when not available or times out when API is down
2. using built-in check script in postgresql image as used in official example template
3. simple / HTTP GET for apidoc


jdobes:
Note: reposcan's healthcheck doesn't need to check database connection because there isn't any opened persistently.

PR #338, opened 2018-06-22
extract duplicated code into super class

tkasparek:
Removing duplicated code from 4 places and putting it into one place.

### DIFF
--- a/webapp/app.py
+++ b/webapp/app.py
@@ -74,6 +74,40 @@ class BaseHandler(tornado.web.RequestHandler):
             data = None
         return data
 
+    @gen.coroutine
+    def handle_post(self, api_endpoint):
+        """Takes care of validation of input and execution of POST methods."""
+        code = 400
+        data = self.get_post_data()
+        if data:
+            try:
+                res = api_endpoint.process_list(data)
+                code = 200
+            except ValidationError as validerr:
+                if validerr.absolute_path:
+                    res = '%s : %s' % (validerr.absolute_path.pop(), validerr.message)
+                else:
+                    res = '%s' % validerr.message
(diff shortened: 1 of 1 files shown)


webapp/app.py
```diff
@@ -143,7 +181,7 @@ def get(self, nevra=None): # pylint: disable=arguments-differ
         yield self.flush()
 
 
-class UpdatesHandlerPost(BaseHandler):
+class UpdatesHandlerPost(PostHandler):
     """Handler for processing /updates POST requests."""
 
     @gen.coroutine
```
grinrag: Take a look at http://www.tornadoweb.org/en/stable/gen.html#tornado.gen.coroutine
`From the caller’s perspective, @gen.coroutine is similar to the combination of @return_future and @gen.engine.`
But this method doesn't return future, do it?

grinrag:
Pls checkout to this and look at the code 8e942872adfba3e2657f8d3b308d79533cb62a14

webapp/app.py
grinrag: If you are going to remove this decorator, make sure it works as expected. You can run simple performance test:
`siege -c 100 -r 10 -H "accept: application/json" 'http://127.0.0.1:8080/api/v1/updates POST < pack.json'`

where `pack.json` is a json with list of nevras

tkasparek:
I've moved handle_post() into BaseHandler, thus removed the PostHandler class.

I've also removed the @gen.coroutine from the places where it wasn't needed.

Regarding performance I did not notice any perfromance drop, all of the upstream and refactored code (both with and without @gen.coroutine) are giving me same number of transactions.

PR #342, opened 2018-06-25
display currently running vmaas version

jdobes:
1. In apidoc Swagger UI
2. Using GET request on /api/v1/version/

### DIFF
--- a/webapp/app.py
+++ b/webapp/app.py
@@ -144,6 +144,22 @@ class ApiSpecHandler(BaseHandler):
         yield self.flush()
 
 
+class VersionHandler(BaseHandler):
+    """Handler class providing app version."""
+
+    @gen.coroutine
+    def get(self): # pylint: disable=arguments-differ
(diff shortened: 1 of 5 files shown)


webapp/app.py
```diff
@@ -718,6 +735,7 @@ def __init__(self):
         handlers = [
             (r"/api/v1/monitoring/health/?", HealthHandler),
             (r"/api/v1/apispec/?", ApiSpecHandler),
+            (r"/api/v1/version/?", VersionHandler),
```
grinrag: should this really be in `/api/v1` endpoint? maybe just `/version` ?

grinrag: I believe `/api/version` should return 1 for now. This looks like version _of_ api. Maybe in this case will use `/app/version` ?

grinrag: Ok, it's not so critical. `/api/v1/version/` is good as well.

Issue #266, opened 2018-05-23
split /updates API and package information


jdobes:
Currently, there are summary and description fields filled for each package having an update in /updates response. This information should be provide by separate API similar to /errata.

ggainey:
Not sure why, but duplicate repositories seem to be returned.

Example NEVRA: `kernel-2.6.32-131.0.15.el6.x86_64`

Returns:
```json
{
  "package_list": {
    "kernel-2.6.32-131.0.15.el6.x86_64": {
      "summary": "The Linux kernel",
      "description": "The kernel package contains the Linux kernel (vmlinuz), the core of any\nLinux operating system.  The kernel handles the basic functions\nof the operating system: memory allocation, process allocation, device\ninput and output, etc.",
      "repositories": [
        "rhel-6-workstation-rpms",
        "rhel-6-workstation-rpms",
        "rhel-6-server-rpms",
        "rhel-6-server-rpms"
      ]
    }
  }
}
```

DB was loaded with these repos:

```json
[{
  "entitlement_cert": {
    "name": "RHSM-CDN",
    "ca_cert": "..."
   "key": ""
  },
  "products": {
    "Red Hat Enterprise Linux Server": {
      "redhat_eng_product_id": 69,
      "content_sets": {
        "rhel-7-server-rpms": {
          "name": "Red Hat Enterprise Linux 7 Server (RPMs)",
          "baseurl": "https://cdn.redhat.com/content/dist/rhel/server/7/$releasever/$basearch/os/",
          "basearch": ["x86_64"],
          "releasever": ["7Server", "7.4"]
        },
        "rhel-7-server-optional-rpms": {
          "name": "Red Hat Enterprise Linux 7 Server - Optional (RPMs)",
          "baseurl": "https://cdn.redhat.com/content/dist/rhel/server/7/$releasever/$basearch/optional/os/",
          "basearch": ["x86_64"],
          "releasever": ["7Server", "7.4"]
        },
        "rhel-6-server-rpms": {
          "name": "Red Hat Enterprise Linux 6 Server (RPMs)",
          "baseurl": "https://cdn.redhat.com/content/dist/rhel/server/6/$releasever/$basearch/os/",
          "basearch": ["x86_64"],
          "releasever": ["6Server", "6.9"]
        },
        "rhel-6-server-optional-rpms": {
          "name": "Red Hat Enterprise Linux 6 Server - Optional (RPMs)",
[...]

PR #373, opened 2018-07-27
Add v2 updates api

eherget:
This adds a v2 updates api, which no longer returns package summary and description.

### DIFF
--- a/webapp/app.py
+++ b/webapp/app.py
@@ -25,6 +25,8 @@ from dbchange import DBChange
 from utils import init_logging, get_logger
 import gen
 
+# pylint: disable=too-many-lines
+
 VMAAS_VERSION = os.getenv("VMAAS_VERSION", "unknown")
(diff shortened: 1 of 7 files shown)


webapp/app.py
```diff
@@ -849,7 +977,8 @@ def setup_apispec(handlers):
     })
     # Register public API handlers to apispec
     for handler in handlers:
-        if handler[0].startswith(r"/api/v1/"):
+        if handler[0].startswith(r"/api/v1/") \
```
ggainey: We could write this with r"/api/v[12]/" instead of two compares, yeah?

eherget: I tried doing that, but it didn't work.  Spent an hour trying to figure out what the correct regex is.  Gave up and did this.  :(

eherget:
If no Jenkins tests failed, why is it set as failing?

eherget:
This fixed Issue #266

PR #374, opened 2018-07-27
Fix #371 - each log-call is a single line/entry

### LINKED ISSUE #371: VMAAS-81: tracebacks log as many-records, makes monitoring way harder than it needs to be

ggainey:
#371 

Extracted logging-code from webapp/utils.py into logging_utils.py
Added OneLineFormatter to logging_utils
Made reposcan/common/logging.py and webapp/logging_utils.py the same code


eherget:
rerun tests

PR #375, opened 2018-07-30
reposcan: introduce authentication for the management API.

grinrag:
1. We also discussed a possibility to set an environment variable to check is it necessary to authorize a user or not. I'm not sure we need it right now and I can add it later.
2. Github personal access token can have specified permissions. For the authorization, we require only permissions `read:user`

### DIFF
--- a/reposcan/reposcan.py
+++ b/reposcan/reposcan.py
@@ -7,6 +7,7 @@ into specified PostgreSQL database.
 import os
 from multiprocessing.pool import Pool
 import json
+import requests
 
 from apispec import APISpec
 from tornado.ioloop import IOLoop, PeriodicCallback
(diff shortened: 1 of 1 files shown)


ggainey:
OK, so for future readers, the appropriate curl-cmd is ```curl -H "Authorization: token <github token here>" <reposcan-url>```

Also, I had to make my membership in RedHatInsights be 'Public' to get things to work, even if the key specified both read:user and read:org. I'm not sure what's up with that - the key is authenticating as 'me', so it should be able to see 'my' orgs, I would think.

In any event - this is def better than what we have right now, and appears to work (given the caveat) - merging.

jdobes:
added this info also to https://github.com/RedHatInsights/vmaas/wiki/ReposcanApi

PR #389, opened 2018-08-21
basearch and releasever can be None, return empty string in result

jdobes:


jdobes:
When specified `basearch` or `releasever` is requested and repo doesn't have it defined, we try to find it in URL.

Issue #370, opened 2018-07-26
/packages : add ability to return all NEVRAs for a specified Name


ggainey:
Return current output (nevra/repo-list), by building all NEVRAs that share the same Name. prodsec would find this useful

jsvob:
Yes, please.

It would be most useful if:

* All versions are returned, even those that are not associated with any CVE. VMaaS currently returns only versions that fix a CVE.
    * As a result, this returns nothing `curl -X GET "http://webapp-vmaas-stable.1b13.insights.openshiftapps.com/api/v1/updates/libvirt-1.1.1-29.el7.x86_64" -H  "accept: application/json" | python3 -m json.tool | grep -E "RHBA|RHEA|RHEA-2017:2594-03|libvirt-3.2.0-14.el7_4.3|RHBA-2017:0098-07|libvirt-2.0.0-10.el7_3.4"`
* The results are formatted as entries similar to "#/definitions/UpdatesResponse", so that it contains at least `package`, `repository`, `basearch`.
    * `repository` is useful for building version trees in case we need them
        * e.g. (look at all these `# fix` comments) https://gitlab.cee.redhat.com/insights-open-source/insights-prodsec-rules/blob/master/prodsec/rules/CVE_2017_5715_cpu_virt.py#L103
        * e.g. https://gitlab.cee.redhat.com/insights-open-source/insights-prodsec-rules/blob/master/prodsec/rules/CVE_2017_5461_nss.py
    * `basearch` is useful in case there are packages that have different Release strings for different architectures - e.g. 'ImageMagick-6.7.8.9-10.ael7b.ppc64le' vs. 'ImageMagick-6.7.8.9-10.el7.x86_64' 

Relevant mailing list thread: http://post-office.corp.redhat.com/archives/vmaas-devel-list/2018-August/msg00017.html

jsvob:
Another example how this would help security rules, if we are to depend on VMaaS:

`dracut` for RHEL 7 never once received a CVE fix (always updated as part of RHEA or RHBA without an associated CVE). So the current VMaaS doesn't return anything about `dracut` in the `/updates` endpoint, understandably.

And yet three RHBAs [1] did fix `dracut`'s behavior where `dracut` didn't correctly apply the CVE-2017-5715 microcode fix on specific AMD CPUs. So we detect `dracut` version in the CVE-2017-5715 rule [2]. To do that, we had to list all released `dracut` NVRs and decide which are/are not vulnerable.

[...]

Issue #369, opened 2018-07-26
/packages : add ability to include 'released-by' errata-id


ggainey:
Either as part of base /packages response, or perhaps as a specific sub-API? prodsec could use this - would also allow us to specify 'release date' (erratum's release-date')

ggainey:
Addressed by #391

PR #391, opened 2018-08-27
add package tree generation and download to reposcan

eherget:


eherget:
run tests

eherget:
This should address issues #369 and #370

eherget:
Note that using the apispec link to "Try it out" for pkgtree/download fails because the browser can't format 100MB of content fast enough.  Use `wget http://localhost:8081/api/v1/pkgtree/download` to test

reposcan/pkgtree.py
```diff
@@ -0,0 +1,204 @@
...
+            for errata_id, cve_id in cursor:
+                self.erratadata[errata_id]['data'].setdefault('cve_list', []).append(self.cvename[cve_id])
+
+    def load_packages(self):
+        """Load the packages info"""
+        with self._named_cursor() as cursor:
+            cursor.execute("""select p.id, p.name_id, p.evr_id, p.arch_id
+                                from package p
+                          inner join evr on p.evr_id = evr.id
+                            order by evr.evr""")
```
jsvob: @eherget Does `order by evr.evr` ensure the epoch-version-release are sorted the same way the rpm library used in yum sorts them?

Or put another way (because I shouldn't need to care about the implementation), is the pkgtree output guaranteed to have the NEVRAs ordered the same way as the rpm library orders them?

Reason I'm asking is that I **don't** want to re-use the following code that we have to have in our current (overly complex) version tree tooling.

```

import rpm  # yum install rpmdevtools is necessary for this

def sort_nevra(unsorted_nevra_list):
    """
    Sorts the items based on the name-version-release string using the rpm library,
    the same way rpmdev-sort does it, which is the same way yum sorts name-version-release strings.
    """

    class NevrVer(object):
        # taken and modified from:
        #
        # rpmdev-sort -- sort rpms from standard input
        #
        # Copyright (c) 2010 Dan Horak
        #               2010-2013 Ville Skyttä
        #
        # This program is free software; you can redistribute it and/or modify
        # it under the terms of the GNU General Public License as published by
        # the Free Software Foundation; either version 2 of the License, or
        # (at your option) any later version.
        #
        # This program is distributed in the hope that it will be useful,
        # but WITHOUT ANY WARRANTY; without even the implied warranty of
        # MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
        # GNU General Public License for more details.
        #
        # You should have received a copy of the GNU General Public License
        # along with this program; if not, write to the Free Software
        # Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
        ver = None
        n = None
        e = None
        v = None
        r = None

        def __init__(self, ver):
            self.ver = ver
            self.e = ""
            self.n, self.v, self.r = ver.name, ver.version, ver.release

        def __cmp__(self, other):
            if self.n > other.n:
                return 1
            elif self.n < other.n:
                return -1
            return rpm.labelCompare((self.e, self.v, self.r),
                                    (other.e, other.v, other.r))

        def __lt__(self, other):
            return self.__cmp__(other) < 0

        def __repr__(self):
            return self.ver.nvr_string

    sorted_nevra_list_tmp = []
    for ver in unsorted_nevra_list:
        arr = []
        arr.append(ver)
        arr.append(NevrVer(ver))
        sorted_nevra_list_tmp.append(arr)
    sorted_nevra_list_tmp.sort(key=lambda x: x[1])
    sorted_nevra_list = [x[0] for x in sorted_nevra_list_tmp]

    return sorted_nevra_list

```

eherget: The `order by evr.evr` makes sure that for a given package name, the list of packages/nevras is ordered.  However, the set of package names are not necessarily guaranteed to be ordered.

As for whether the list of packages/nevras for a given package name is ordered the way the rpm library orders them, I'm not certain and will reach out to someone on my team (@grinrag ) to find out.

One thing I see by examining the code you posted is that the posted code does not take epoch into account (its set to "" and doesn't appear to get modified anywhere).  Our sorting of evr compares the epoch from the package/nevra, then the version, then the release.

eherget: So I've learned that the way we do comparison of EVR came from the implementation found in Sat5.  We are not sure where that implementation came from.  Hopefully by understanding what our code does, you might be able to determine if its the same as rpm library?

Below is a snippet of db code that builds an array of evr_array_item (we defined) items from the components of the version and release values.  These arrays and the epoch are put into a column of type evr_t (we defined), and then when we query the database we sort on this column.

When parsing a version or release where some component has both letters and numbers in it, the array of evr_array_items looks a bit strange.  Here's a couple samples of the arrays produced by the parsing code below (rpmver_array):

1.2.3 -> [(1, NULL), (2, NULL), (3, NULL)]
d38b4796  ->  [(0, 'd'), (38, NULL), (0, 'b'), (4796, NULL)]

570c8cd3 -> [(570, NULL), (0, 'c'), (8, NULL), (0, 'cd'), (3, NULL)]

svn16832.2.16b -> [(0, 'svn'), (16832, NULL), (2, NULL), (16, NULL), (0, 'b')]

```
-- -----------------------------------------------------
-- evr type
-- represents version and release as arrays of parsed components
-- of proper type (numeric or varchar)
-- this allow us to use standard array operation for = > <
-- and most importantly sorting
-- -----------------------------------------------------
create type evr_array_item as (
        n       NUMERIC,
        s       TEXT
);

create type evr_t as (
        epoch INT,
        version evr_array_item[],
        release evr_array_item[]
);

create or replace FUNCTION rpmver_array (string1 IN VARCHAR)
RETURNS evr_array_item[] as $$
declare
    str1 VARCHAR := string1;
    digits VARCHAR(10) := '0123456789';
    lc_alpha VARCHAR(27) := 'abcdefghijklmnopqrstuvwxyz';
    uc_alpha VARCHAR(27) := 'ABCDEFGHIJKLMNOPQRSTUVWXYZ';
    alpha VARCHAR(54) := lc_alpha || uc_alpha;
    one VARCHAR;
    isnum BOOLEAN;
    ver_array evr_array_item[] := ARRAY[]::evr_array_item[];
BEGIN
    if str1 is NULL
    then
        RAISE EXCEPTION 'VALUE_ERROR.';
    end if;

    one := str1;
    <<segment_loop>>
    while one <> ''
    loop
        declare
            segm1 VARCHAR;
            segm1_n NUMERIC := 0;
        begin
            -- Throw out all non-alphanum characters
            while one <> '' and not isalphanum(one)
            loop
                one := substr(one, 2);
            end loop;
            str1 := one;
            if str1 <> '' and isdigit(str1)
            then
                str1 := ltrim(str1, digits);
                isnum := true;
            else
                str1 := ltrim(str1, alpha);
                isnum := false;
            end if;
            if str1 <> ''
            then segm1 := substr(one, 1, length(one) - length(str1));
            else segm1 := one;
            end if;

            if segm1 = '' then return ver_array; end if; /* arbitrary */
            if isnum
            then
                segm1 := ltrim(segm1, '0');
                if segm1 <> '' then segm1_n := segm1::numeric; end if;
                segm1 := NULL;
            else
            end if;
            ver_array := array_append(ver_array, (segm1_n, segm1)::evr_array_item);
            one := str1;
        end;
    end loop segment_loop;

    return ver_array;
END ;
$$ language 'plpgsql';

-- -----------------------------------------------------
-- Table vmaas.evr
-- -----------------------------------------------------
CREATE TABLE IF NOT EXISTS evr (
  id SERIAL,
  epoch TEXT NOT NULL, CHECK (NOT empty(epoch)),
  version TEXT NOT NULL, CHECK (NOT empty(version)),
  release TEXT NOT NULL, CHECK (NOT empty(release)),
  evr evr_t NOT NULL,
  UNIQUE (epoch, version, release),
  PRIMARY KEY (id)
)TABLESPACE pg_default;
```

eherget:
So I think this means the ordering should be the same for a given package, assuming the evr-compare in rpmdb hasn't changed.

jsvob:
@eherget @ggainey Thank you for the detailed answer. It is really good that it sorts the NEVRAs that way, it makes our code much easier.

jsvob:
Also sorry for the late answer. I thought I'd manage to test the data thoroughly earlier. I just did and pkgtree is exactly what we were looking for!

reposcan/pkgtree.py
jsvob: > One thing I see by examining the code you posted is that the posted code does not take epoch into account (its set to "" and doesn't appear to get modified anywhere). Our sorting of evr compares the epoch from the package/nevra, then the version, then the release.

That is correct and so far, intentional - the data sources we currently use don't provide the epoch information and the packages we had to write detection for in the last 3 years didn't have such a versioning that omitting Epoch would break it. Especially now that we (will) take the N(E)VR(A) ordering directly from VMaaS, the only instance where omitting Epoch on our part would be a problem is the case where (Name,Version,Release) is exactly the same string, and given RH's tendency to bump Release every time, that's probably not going to happen. But yes, if we were doing the sort of thing you do (detection for everything, not just hand-picked CSAws), we might have a problem with the code I posted.

PR #397, opened 2018-09-12
add comments on environment variable for the database container.

grinrag:
Environment variables with username and password are used to create a new account in the DB.


grinrag:
That was not really clear that the user from env variable will be created.

grinrag:
rerun tests

PR #406, opened 2018-09-21
Store module metadata in the VMaaS DB

tkasparek:
Downloads, parses and stores all relevant info regarding modules into the VMaaS database.


ggainey:
Code looks good, results of running against a rhel8 repo look good. Test-failures appear to be unrelated - I've seen this before, they might just go away on re-run

ggainey:
rerun tests

Issue #402, opened 2018-09-19
Teach pkgtree to gzip output


ggainey:
pkgtree results for  systems with Many Repos can be LARGE. Teach pkgtree to gzip results

eherget:
When we gzip the pkgtree file, it goes from about 3.2Gb to 90Mb using default compression.

ggainey:
fixed by #407

Issue #394, opened 2018-08-30
pkgtree - add ability to set environment variable to generate pkg_tree.json with un-prettified content to save a little space


eherget:

eherget:
This is added in #407

PR #407, opened 2018-09-24
Gzip pkgtree

eherget:
This addresses issue #402 .  In addition to gzip'ing the pkgtree file, it also chunks the reading of the file when downloaded.

This also addresses issue #394 allowing an environment variable to specify the pkgtree indent (ie pretty output).  It defaults to 0 to minimize the size of the pkgtree file.


ggainey:
works in my env, jenkins-failure is another 'unrelated to current commit" - merged

PR #415, opened 2018-10-03
fix #410 - split reposcan and websocket server

### LINKED ISSUE #410: move websocket server into separate service
reposcan API is exposed to outside networks but read/write to websocket should be possible only between internal services in either vmaas or vulnerability-engine

jdobes:
websocket server should be used for internal communication between nodes, should not be exposed as public API handler (as part of reposcan), this is also security risk

this PR is moving websocket server to a separate service


ggainey:
test failures due to new cvss_metrics in db - expected.

PR #418, opened 2018-10-10
Fix #417 - make pkgtree/db-export 'keep copies' to be env-vars instea…

### LINKED ISSUE #417: reposcan/exporter - "num dump copies to keep" needs to be configurable
Currently KEEP_COPIES is hardcoded in reposcan/exporter.py - needs to be moved to reposcan.env, and needs to default to 3 not 10

ggainey:
…d of hardcoded


jdobes:
env variables added also into openshift template

PR #422, opened 2018-10-19
Fix #411 - adding "exported" to the dbchange api docs

### LINKED ISSUE #411: update dbchange api docs
the apispec docs for dbchange api is missing the "exported" field

tlestach:


tlestach:
Thank you @jdobes !

PR #423, opened 2018-10-22
Webapp unit tests

psegedy:
Add unit tests for webapp, using pytest framework.
Tests need data from vmaas DB `webapp/test/data/cache.yml`. With every change in DB schema, please update this file following https://url.corp.redhat.com/a10821a

After build in Travis, coverage results will be uploaded to Codecov.io. Codecov.io will then update pull request status with actual code coverage and mark build as failed if coverage decrease.

To use codecov it may be necessary to allow this service, if it's not allowed in RedHatInsights organization.


ggainey:
Looks good! altho codecov fails when running locally - I assume just setting $CODECOV_TOKEN would be sufficient, if one wanted to do so.

PR #425, opened 2018-11-01
add helper script to remove all resource requirements from all deploy…

jdobes:
…ment configs in current project

Our openshift templates are using resource requests/limits for each service, this is good for production but when deploying in local openshift cluster like minishift, you may not be able (and you typically don't need to) to allocate so much resources, this script will remove all resource requirements from currently selected project.


jdobes:
added script to switch container to devel mode (by running sleep in it)

jdobes:
closing in favor of https://github.com/RedHatInsights/vulnerability-engine/pull/42

PR #427, opened 2018-11-05
Reposcan unittests

psegedy:
Add tests in pytest framework for reposcan.
- test common module
- test database module, testing that data is stored in DB

Note: travis build will fail, because of missing webapp tests. Merge first https://github.com/RedHatInsights/vmaas/pull/423, then this PR


psegedy:
@ggainey vmaas-ci was failing because there were no tests for webapp, thus pytest exited with exit code 5 `no tests ran in 0.11 seconds`.

Jenkins build is failing, because it doesn't have github token for authetication (we are making requests to GH to get state of issues). This is caused by moving tests under IQE. IQE is using https://github.com/rochacbruno/dynaconf to provide configuration files and there is some bug caused by Dynaconf. I'm sure that @mkoura know more about this.

After resolving conflicts, vmaas-ci is passing.

mkoura:
@psegedy FYI the token problem is already solved.

ggainey:
Fixed a rebase-issue and merged to master - thanks!

PR #431, opened 2018-12-05
Raw-download is ignoring Content-Encoding header in transfers

ggainey:


eherget:
Looks good.  Works when I test it.  Merging...

PR #436, opened 2018-12-14
Fixed spec to use objects instead of hardcoded

psav:


jdobes:
is it possible to change examples of keys instead of additionalProp1, additionalProp2, additionalProp3 etc.?

PR #441, opened 2019-01-15
prometheus probe tweaks

ggainey:
- Added probe to v2/updates (oops)
 - Added Histogram to cve/errata/repos (in addition to /updates)
 - Renamed invocations-counter to make it more clear when collected by the prom-server


jdobes:
rerun tests

PR #455, opened 2019-02-20
fix import of erratas with same packe multiple times in it

tkasparek:


ggainey:
Tested with clean DB and a couple thousand repos - accepting

PR #458, opened 2019-02-27
process modular updates

tkasparek:
This PR addresses modularity updates and modular erratas.
There's a new parameter added to the POST /updates API called modules_list (see SwaggerUI), which serves as constraint to the updates search.
Without "modules_list" the /updates API returns every possible update, with no regards to enabled modules (it will offer package from other stream as an update if it has higher NEVRA).
With "modules_list" parameter present, but empty, the /updates API expects that no modules are enabled and will discard all modular erratas.
When "modules_list" is present and NOT empty, /updates API will show modular updates coming only from modules in the "modules_list" and all non-modular updates as well.

### DIFF
--- a/reposcan/exporter.py
+++ b/reposcan/exporter.py
@@ -336,25 +336,16 @@ class DataDump:
     def dump_modules(self, dump):
         """Select module information"""
         with self._named_cursor() as cursor:
-            if self.errata_ids:
-                cursor.execute("""select pkg_id,
-                                         module_stream_id as module_id
-                                    from pkg_errata
-                                   where module_stream_id is not null
-                                     and errata_id in %s
-                                   union
-                                  select pkg_id,
-                                         stream_id as module_id
-                                    from module_rpm_artifact
-                            """, [tuple(self.errata_ids)])
-            else:
-                cursor.execute("""select pkg_id,
-                                         stream_id as module_id
-                                    from module_rpm_artifact
-                               """)
+            cursor.execute("""select pkg_id,
+                                     errata_id,
+                                     module_stream_id as module_id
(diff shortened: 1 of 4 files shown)


reposcan/exporter.py
```diff
@@ -336,25 +336,16 @@ def dump_cves(self, dump):
     def dump_modules(self, dump):
         """Select module information"""
         with self._named_cursor() as cursor:
-            if self.errata_ids:
```
jdobes: this if is still needed, select will fail in case of no errata in repo

webapp/updates.py
```diff
@@ -261,8 +271,10 @@ def _build_nevra(self, update_pkg_id):
         arch = self.db_cache.id2arch[arch_id]
         return join_packagename(name, epoch, ver, rel, arch)
 
-    def _process_updates(self, packages_to_process, api_version, available_repo_ids, repo_ids_key, response):
+    def _process_updates(self, packages_to_process, api_version, available_repo_ids,
+                         repo_ids_key, response, module_ids):
         # pylint: disable=too-many-branches
+        module_filter = module_ids is not None
```
jdobes: why is the variable here when it's used only on L335?

tkasparek: if it would be defined on L335 that would mean that the evaluation of the expression is done multiple times (inside two for cycles) this way, this expression is evaluated just once

reposcan/exporter.py
tkasparek: Will change this

webapp/updates.py
jdobes: kk

PR #463, opened 2019-03-06
VMAAS-367 - always clean data in temp

jdobes:


jdobes:
pylint issues fixed by #462

PR #464, opened 2019-03-07
add unit tests for modularity

tkasparek:


webapp/test/test_modularity.py
```diff
@@ -0,0 +1,146 @@
...
+        #   postgresql-9.6.10-1.module+el8+2470+d1bafa0e.x86_64.rpm
+        #   postgresql-10.6-1.module+el8+2469+5ecd5aae.x86_64.rpm
+        updates = self.updates_api.process_list(2, self.gen_pkg_json(PKG_NEVRAS[0]))
+        assert updates
+        available_updates = updates['update_list'][PKG_NEVRAS[0]]['available_updates']
+        assert len(available_updates) == 2
+        expected_updates = {PKG_NEVRAS[1], PKG_NEVRAS[3]}
+        assert expected_updates == {rec['package'] for rec in available_updates}
+
+        # Test with modularity info provided, should find one update:
```
psegedy: It would be better to split it into to tests. Now, if testing without modularity fail, test with modularity won't be executed.
You can parametrize tests in pytest with `@pytest.mark.parametrize` decorator [1]

In this case it could be something like:
```python
# name, module, expected
OLD_OLD = [
    ('without_modularity', None, {PKG_NEVRAS[1], PKG_NEVRAS[3]}),
    ('with_modularity', MODULE_JSONS[0], {PKG_NEVRAS[1]})
]


@pytest.mark.parametrize("module", OLD_OLD, ids=[n[0] for n in OLD_OLD])
def test_old_pkg_old_module(self, module):
...
updates = self.updates_api.process_list(2, self.gen_pkg_json(PKG_NEVRAS[0], [module[1]]))
...
expected_updates = module[2]
assert expected_updates == {rec['package'] for rec in available_updates}
...

```
...and similarly for other tests

[1] https://docs.pytest.org/en/latest/parametrize.html

Issue #456, opened 2019-02-25
NIST read failing with decode-error


ggainey:
```
__main__: [ERROR] Internal server error <-9223363284117796852>
'Traceback (most recent call last):
  File "/reposcan/reposcan.py", line 687, in run_task
    controller.store()
  File "/reposcan/nistcve/cve_controller.py", line 133, in store
    repo.load_json()
  File "/reposcan/nistcve/cverepo.py", line 37, in load_json
    self.json = CveJson(self.json_tmp())
 File "/reposcan/nistcve/cvejson.py", line 15, in __init__
    self.data = json.load(fde)\n  File "/opt/rh/rh-python36/root/usr/lib64/python3.6/json/__init__.py", line 296, in load
    return loads(fp.read(),
  File "/opt/rh/rh-python36/root/usr/lib64/python3.6/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: \'ascii\' codec can\'t decode byte 0xe2 in position 1713354: ordinal not in range(128)'|
```

ggainey:
This blocks updating info, need to fix

jdobes:
fixed by #456

PR #467, opened 2019-03-13
#456 - recognize that NIST-json can be utf-8

ggainey:
Addresses #456


jdobes:
verified that cves sync task succeeds

PR #474, opened 2019-04-04
 reduce number of intermediate Docker images after code change 

tkasparek:
Previously we were adding our code before running e.g. pip install or other things which are changing less often and have bigger layers than our code.
I've restructured the Dockerfiles in a way that things that are changing the least and take the biggest amount of time are run first, and our code is added at last possible place so we won't run whole `pip install --upgrade && pip install $LOT_OF_STUFF` after one line of code gets changed.

### DIFF
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -1,5 +1,28 @@
 #!/usr/bin/bash
 
+rc=0
+
+for dockerfile in Dockerfile-database Dockerfile-reposcan Dockerfile-webapp Dockerfile-websocket
+do
+    if [ ! -f "$dockerfile" ]; then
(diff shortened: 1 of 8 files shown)


run_tests.sh
```diff
@@ -1,5 +1,26 @@
 #!/usr/bin/bash
 
+for dockerfile in Dockerfile-reposcan Dockerfile-webapp Dockerfile-websocket
```
jdobes: Dockerfile-database is missing

jdobes: `Dockerfile-database and Dockerfile-database.rhel7 are too different!`

it's because of hardcoded PG 10 value in test, please add a commit switching database to PG 10

also make sure run_tests.sh fails if error is detected

PR #477, opened 2019-04-25
add probes for /updates hot cache hits/misses

tkasparek:


ggainey:
So I am not sure what in the test-harness is failing, but this code works Just Fine in a clean environment. I am accepting the PR, with the caveat that we *may* need a quick fix first thing in the AM.

tkasparek:
@ggainey as far as I know there's an issue in Jenkins tets and QA is aware of those and working on resolving those. As you pointed out, the code works and the issue is somewhere else.

PR #481, opened 2019-04-29
Fix traceback when importing repos w/modules and without at the same …

ggainey:
…time

Import creates known-package-map at startup and assumes nothing can change it inside a single run.
Modularity-related code violates that assumption.

Teach package_store to not rely on __init__-time processing so much.


psegedy:
Jenkins is failing because `ggainey:reposcan_modularity_tb` branch is missing needed Dockerfile - `Dockerfile-webapp-qe`

PR #484, opened 2019-05-08
add reposcan exporter unit tests

eherget:
Add tests for exporter.py.

I noticed a likely issue in exporter exporting modules data.  I'll open an issue for that.

### DIFF
--- a/reposcan/test/conftest.py
+++ b/reposcan/test/conftest.py
@@ -1,0 +1,48 @@
+"""Pytest configuration for database tests."""
+
+import os
+
+import psycopg2
+import pytest
+import testing.postgresql
+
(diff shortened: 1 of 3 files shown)


tkasparek:
I think you've accidentally added some cache file, look at ` reposcan/test_data/exporter/exporter_test_data.sql~` and if it's some kind of temptile, could you please delete it and force push you changes without it?

eherget:
Done.  Sorry about that.  emacs creates files with trailing '~' character in case of crash.

Josca:
@eherget yes, it's common feature of many text editors. Maybe we could add `*~` item to `.gitignore`, what do you mean?

PR #486, opened 2019-05-08
Fix unique-constraint-violation error when importing repos with modul…

ggainey:
…es with errata

- Remove pkg_errata_pkg_id_errata_id_key
- Add pkg_errata_pkgid_errataid and pkg_errata_pkgid_streamid_errataid to
  a) add module_stream_id to uniqueness while b) correctly dealing with its null-ness


eherget:
Looks good.  We should double check with tkasparek and make sure the intended functionality is maintained.

PR #489, opened 2019-05-15
create API for getting repositories that changed after certain date

tkasparek:


jdobes:
`curl http://127.0.0.1:8080/api/v1/repochange/999`
...
returns stuff, is this expected input format?

jdobes:
there is a `parse_datetime` function in utils.py used for modified_since filtering, please consider using common function for all datetime parsing

Issue #453, opened 2019-02-18
Store information about source packages


Allda:
Red Hat security metadata usually provides information about affected packages in one of these 2 formats.

* RPM packages (NEVRA - name-epoch:version-release.arch)
* Source packages (NEVR - name-epoch:version-release)
Unfortunately when a user uses multiple security data sources and each provides data in the different format it is always difficult to map package from one format to another.

It would be great if VMaaS could provide not only information about binary packages but also information about source packages.

ggainey:
Minor nit - SRPM info in primary.xml is only NVR (for historical reasons that escape me at the moment)

The goal here would be to provide a /srpm API, which you could hand an SRPM NVR, and get back a list of the package NEVRAs that think they were built from from that SRPM - does that match your expectation?

Allda:
I included epoch in source package because of the mapping file includes it as well.
https://www.redhat.com/security/data/metrics/cvemap.xml

But yeah, the general idea is to provide a mapping between rpm and srpm. You can provide another /srpm API, but it would be even better to include srpm in already created APIs (/advisory  or /cve or /packages) because in this case we could do simple API query instead several queries.

tlestach:
@Allda, would you describe your use case, please?
What existing API would you need to add the srpm information? Is there any new API you'd need to introduce?

Allda:
@tlestach We are currently using v1/errata/<advisory> API to get a list of rpms which are included in that advisory and then we have to map it to srpm using internal Red Hat's tools. It would be great if srpm information would be included in v1/errata/<advisory> so we can do mapping in just single call.

jdobes:
@Allda We're preparing a new version providing SRPM mappings in all v1/errata, v1/cves and v1/packages APIs.

This is now deployed in stage environment, please check if these changes match your expectations.

https://apidoc-vmaas-stage.1b13.insights.openshiftapps.com

Example:
`curl -X GET "https://webapp-vmaas-stage.1b13.insights.openshiftapps.com/api/v1/errata/RHSA-2018:0512" -H  "accept: application/json"`
```
...
"source_package_list": [
    "kernel-2.6.32-696.23.1.el6.src"
],
...
```

Allda:
@jdobes Thanks for example with included srpm, unfortunately with this output we are still not able to tell what it a relation between rpm and srpm. We would need a mapping between those 2 lists.

Something like:
```
srpm_mapping: {
[...]

PR #492, opened 2019-05-17
SRPM parse and store

Josca:


reposcan/database/package_store.py
```diff
@@ -102,41 +60,63 @@ def _populate_evrs(self, unique_evrs):
...
+        for name, epoch, version, release in srpms:
+            evr_id = self.evr_map[(epoch, version, release)]
+            name_id = self.package_name_map[name]
+            unique_packages[(name_id, evr_id)] = (name_id, evr_id)
+        to_import = []
+        for name_id, evr_id in unique_packages:
+            if (name_id, evr_id) not in self.srpm_map:
+                to_import.append(unique_packages[(name_id, evr_id)])
+
+        self.logger.debug("Packages to import: %d", len(to_import))
```
jdobes: Change to "Source packages to import"?

jdobes:
I'm still considering if it's better idea to reuse `package` table for srpms as well and not to create `srpm` table. Basically only thing preventing us to do that now is unique not null constraint on checksum column so it'd need to relax schema a bit.

It'd make expanding stored information about srpms easier. This PR only stores info about srpm we can get from regular repositories. But there are also srpm repositories storing source packages  (for example https://dl.fedoraproject.org/pub/fedora/linux/releases/30/Server/source/tree/) itself and containing all missing info about source packages (description, summary, checksum, in which repo srpm is etc.). These information may be required even if #453 doesn't mention it yet.

I think we should consider source packages as regular packages, set all attributes, associate with repos, errata etc. In this case there is no need to create new srpm table, just modify the package table to allow insert/update in two steps:
1. what this PR implements - store only a stanza of source package (missing checksum, description, not associated with any repo etc.) and link regular packages rows to srpm rows (package table is self-referencing itself)
2. while syncing source RPM repository update source package rows with additional information

*Note that order of these two steps isn't guaranted

jdobes:
if #496 is merged then re-using package table for srpms is possible

reposcan/database/package_store.py
```diff
@@ -80,7 +80,6 @@ def _populate_package_names(self, unique_names):
         cur = self.conn.cursor()
         self.logger.debug("Unique package names in repository: %d", len(unique_names))
         to_import = []
-        self.package_name_map = self._prepare_table_map(["name"], "package_name")
```
jdobes: can't simply remove it, it's here intentionally - see c6e89e3d7727ed81409e01c3b4619bad4de3fd46

it's not optimal and the issue is tracked here - https://github.com/RedHatInsights/vmaas/issues/494

Josca: Sorry. I thought it's redundant and tests passed after removing. I see it's good that review is done by you.

jdobes: actually I think it may be deleted, it's beacuse of modules_store.py inserting to the package_name table but whole modules_store.py is unused

Josca: :smile: , ok, cool

tlestach:
> I'm still considering if it's better idea to reuse `package` table for srpms as well and not to create `srpm` table. Basically only thing preventing us to do that now is unique not null constraint on checksum column so it'd need to relax schema a bit.

I understand @jdobes's point and see the 'package' table exactly covers the needs to store a srpm, but let me mention two arguments, why I'd consider storing srpms in a separate table:
* I believe the table 'package' is already quite large (can someone check how many entries are there?)
- my concern is to slow down package related queries, if we imported also srpm entries into the same table
- srpms need to reference rpm entries built from the same source rpm - and creating this reference within on table sounds odd, doesn't it?

psegedy:
Jenkins failed with
```
__main__: [ERROR] Internal server error <-9223363290297850382>'Traceback (most recent call last):\n File "/reposcan/reposcan.py", line 530, in run_task\n export_data(DUMP)\n File "/reposcan/exporter.py", line 386, in main\n data.dump()\n File "/reposcan/exporter.py", line 42, in dump\n self.dump_package_details(dump)\n File "/reposcan/exporter.py", line 133, in dump_package_details\n """, [tuple(self.packagename_ids)])\npsycopg2.errors.UndefinedColumn: column "srpm_id" does not exist\nLINE 1: ...OUT HOLD FOR select id, name_id, evr_id, arch_id, srpm_id, s...\n ^\nHINT: Perhaps you meant to reference the column "package.arch_id".\n'|
```

reposcan/database/package_store.py
```diff
@@ -128,11 +140,25 @@ def _associate_packages(self, package_ids, repo_id):
...
         Import all packages from repository into all related DB tables.
         """
-        self.logger.debug("Syncing %d packages.", len(packages))
+        source_packages = self._get_source_packages(packages)
+        self.logger.debug("Syncing %d packages.", len(packages) + len(source_packages))
+        self._populate_dependent_tables(packages)
+        source_package_ids = self._populate_packages(source_packages)
         package_ids = self._populate_packages(packages)
-        self._associate_packages(package_ids, repo_id)
+        self._associate_packages(source_package_ids + package_ids, repo_id)
```
jdobes: source_package_ids shouldn't be associated with this repo

reposcan/database/package_store.py
```diff
@@ -128,11 +140,25 @@ def _associate_packages(self, package_ids, repo_id):
...
+
     def store(self, repo_id, packages):
         """
         Import all packages from repository into all related DB tables.
         """
-        self.logger.debug("Syncing %d packages.", len(packages))
+        source_packages = self._get_source_packages(packages)
+        self.logger.debug("Syncing %d packages.", len(packages) + len(source_packages))
+        self._populate_dependent_tables(packages)
+        source_package_ids = self._populate_packages(source_packages)
```
jdobes: this seems crashing with KeyError:
```
  File "/reposcan/reposcan.py", line 661, in run_task
    repository_controller.store()
  File "/reposcan/repodata/repository_controller.py", line 264, in store
    self.repo_store.store(repository)
  File "/reposcan/database/repository_store.py", line 167, in store
    self.package_store.store(repo_id, repository.list_packages())
  File "/reposcan/database/package_store.py", line 161, in store
    source_package_ids = self._populate_packages(source_packages)
  File "/reposcan/database/package_store.py", line 90, in _populate_packages
    evr_id = self.evr_map[(pkg["epoch"], pkg["ver"], pkg["rel"])]
KeyError: ('0', '1.1.2', '3.el5_1.2')
```

database/vmaas_db_postgresql.sql
```diff
@@ -399,6 +399,7 @@ CREATE TABLE IF NOT EXISTS package (
   arch_id INT NOT NULL,
   summary TEXT NULL, CHECK (NOT empty(summary)),
   description TEXT NULL, CHECK (NOT empty(description)),
+  srpm_id INT NULL,
```
jdobes: please rename to source_package_id, it's referencing package table

reposcan/database/package_store.py
Josca: Ok, resolved using last commits.

PR #510, opened 2019-06-06
packages sorting with ~ and ^ notation

Josca:
added support for ~ and ^ notation (sorting)
- sql function rpmver_array replaced with python one
- added several tests for related functions


eherget:
This looks good.  Merging.

PR #511, opened 2019-06-06
removed github-fetch-pullrequest script, packed to PyPI

Josca:
I created python package [github-fetch-pullrequest](https://github.com/Josca/github-fetch-pullrequest).
Let's use it instead of storing duplicit code in each repo.


eherget:
Just a note for the readme.md on your github-fetch-pullrequest github repo... I needed to use pip3 to install it so it is run with python3.  The default caused it to install and run with python2.7.

PR #520, opened 2019-06-17
Add modules to pkgtree output

eherget:
This is to address Issue # 506 which requested that module information be added to pkgtree output.

### DIFF
--- a/reposcan/database/modules_store.py
+++ b/reposcan/database/modules_store.py
@@ -2,13 +2,32 @@
 Module containing classes for fetching/importing modules from/into database.
 """
 
+import re
+
 from psycopg2.extras import execute_values
 
 from database.object_store import ObjectStore
 
-
 class ModulesStore(ObjectStore):
     """Class providing interface for storing modules and related info."""
+
+    def __init__(self):
+        super().__init__()
+        self.nevra_re = re.compile(r'(.*)-(([0-9]+):)?([^-]+)-([^-]+)\.([a-z0-9_]+)')
+
+    def _split_packagename(self, filename):
+        """Split rpm name (incl. epoch) to NEVRA components."""
+        if filename[-4:] == '.rpm':
+            filename = filename[:-4]
+
+        match = self.nevra_re.match(filename)
+        if not match:
+            return '', '', '', '', ''
+
+        name, _, epoch, version, release, arch = match.groups()
+        if epoch is None:
+            epoch = '0'
+        return name, epoch, version, release, arch
 
     def _populate_modules(self, repo_id, modules):
         cur = self.conn.cursor()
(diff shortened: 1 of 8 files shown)


eherget:
While implementing this, I noticed a couple things I have questions about.  See the following output from pkgtree:

```
{
"nevra": "apache-commons-codec-1.11-3.module+el8+2452+b359bfcd.noarch",
"repositories": [
{
"lable": "rhel-8-for-x86_64-appstream-rpms",
"name": "Red Hat Enterprise Linux 8 for x86_64 - AppStream (RPMs)",
"arch": "",
"releasever": "8.0",
"revision": "2019-06-13T13:07:45+00:00",
"module_name": "maven",
"module_stream": "3.5"
},
{
"lable": "rhel-8-for-x86_64-appstream-rpms",
"name": "Red Hat Enterprise Linux 8 for x86_64 - AppStream (RPMs)",
"arch": "",
"releasever": "8",
"revision": "2019-06-13T13:07:45+00:00",
"module_name": "maven",
"module_stream": "3.5"
}
]
}
```

Two questions:
1) why is the arch not included in these repositories even though the name suggests they are specific to x86_64 in this example?  It seems all repos with modules do not have arch defined.
2) why is there both '8' and '8.0' for releasever?

reposcan/database/modules_store.py
```diff
@@ -2,14 +2,33 @@
...
 
-
 class ModulesStore(ObjectStore):
     """Class providing interface for storing modules and related info."""
 
+    def __init__(self):
+        super().__init__()
+        self.nevra_re = re.compile(r'(.*)-(([0-9]+):)?([^-]+)-([^-]+)\.([a-z0-9_]+)')
+
+    def _split_packagename(self, filename):
```
Josca: We have function for that already. It's `common.rpm.parse_rpm_name`. Could you use it instead of this, please?

reposcan/database/modules_store.py
```diff
@@ -87,9 +106,119 @@ def _populate_streams(self, modules):
...
+        for module in modules:
+            for profile in module['profiles']:
+                profiles.add((module['stream_id'], profile,))
+        if profiles:
+            execute_values(cur,
+                           """select id, stream_id, profile_name from module_profile
+                              inner join (values %s) t(stream_id, profile_name)
+                              using (stream_id, profile_name)
+                           """, list(profiles), page_size=len(profiles))
+            for row in cur.fetchall():
```
Josca: I think `for id, stream_id, profile_name in cur.fetchall():...` would be more clear here. You don't need to use indexing over `row` variable in next two lines.

Josca: The same at line 164.

reposcan/pkgtree.py
```diff
@@ -57,16 +60,19 @@ def dump(self):
...
-        self.load_evr()
-        self.load_arch()
-        self.load_repodata()
-        self.load_cves()
-        self.load_errata()
-        self.associate_cves_to_errata()
-        self.load_packages()
-        self.associate_repos()
-        self.associate_errata()
+        self._load_packagenames()
```
Josca: Yes, exactly, let's make as much methods private as possible :smiley:.

reposcan/repodata/modules.py
```diff
@@ -0,0 +1,61 @@
...
+        for name in module_dict:
+            for stream in module_dict[name]:
+                self.modules.append(module_dict[name][stream])
+
+    @staticmethod
+    def _parse_modulemd_defaults(module_dict, default_list):
+        for parsed in default_list:
+            data = parsed['data']
+            name = data.get('module')
+            default_stream = data.get('stream', None)
```
Josca: Second argument is useless, `None` is default return variable.

reposcan/repodata/modules.py
```diff
@@ -0,0 +1,61 @@
...
+            if default_stream:
+                module_dict[name][default_stream]['default_stream'] = True
+            for stream in data['profiles']:
+                if stream not in module_dict[name]:
+                    continue # this is here due to module defaults referencing non-existing modules
+                default_profile = data['profiles'][stream]
+                if default_profile and default_profile[0] in module_dict[name][stream]['profiles']:
+                    module_dict[name][stream]['profiles'][default_profile[0]]['default_profile'] = True
+
+    def list_modules(self):
```
Josca: That's quite big update of code here. It would be good also to update tests of storing data to database. I mean to extend `test_repository_store.py` or create new test module `test_modules.py`.

eherget: Most of this code "update" was actually code restored from a previous commit that removed it.  Note that in the commit that reverted the commit removing so much code from modules_store.py and modules.py, it not only restored all of that code, but also restored reposcan/repodata/test/test_modules.py and reposcan/test_data/repodata/modules.yaml.  So there are some tests that load and parse yaml.  We should look to add tests to store and retrieve from the database as well.

eherget: Since my task was to add module data to pkgtree output and there is no unit tests for pkgtree, I'll create tests for pkgtree and add to this PR

reposcan/pkgtree.py
```diff
@@ -182,14 +187,22 @@ def add_package_entry(self, pkg_id, name_id, evr_id, arch_id):
...
         self.datadict[name_id].append(self.packagedata[pkg_id]['data'])
 
-    def associate_repos(self):
+    def _associate_repos(self):
         """Add repodata to the package entries"""
         with self._named_cursor() as cursor:
             cursor.execute("select pkg_id, repo_id from pkg_repo")
             for pkg_id, repo_id in cursor:
-                self.packagedata[pkg_id]['data']['repositories'].append(self.repodata[repo_id]['data'])
+                addrepo = {'lable': self.repodata[repo_id]['data']['label'],
```
tkasparek: typo: `lable` instead of `label`

tkasparek: there's a typo, `lable` instead of `label`

eherget:
WIP - TODO
1) Make code handle db exceptions - bring up to speed with VMAAS-334 task
2) make modifications Josef suggested
3) investigate why repos with modules doesn't have basearch
4) create test_pkgtree unit tests
5) fix "lable" to be "label" in pkgtree output

reposcan/pkgtree.py
eherget: I see what I did... it hasn't been there forever.  I introduced it in one of my commits.  No longer could I just copy the "data" value (a dict of values for pkgtree output) to the output because for some repos I need to add module_name and module_stream to this output dict.  Now it builds a new output dict for each repo and copies the label, name, arch, releasever and revision values to it (and adds module_name and module_stream where appropriate)

reposcan/database/modules_store.py
eherget: common.rpm.parse_rpm_name doesn't work identically to _split_packagename.  For example, it throws RPMParseException for 'perl-DBD-Pg-0:3.7.4-2.module+el8+2517+b1471f1c.x86_64'

eherget: It should parse it to:
('perl-DBD-Pg', '0', '3.7.4', '2.module+el8+2517+b1471f1c', 'x86_64')

eherget: On closer inspection, the _split_packagename() and common.rpm.parse_rpm_name() methods are different.  It looks like common.rpm.parse_rpm_name() doesn't correctly parse out the epoch when passed "perl-DBD-Pg-0:3.7.4-2.module+el8+2517+b1471f1c.x86_64.rpm".
It returns ('perl-DBD-Pg', None, '0:3.7.4', '2.module+el8+2517+b1471f1c', 'x86_64')

I'm going to leave the _split_packagename() method in place for now.  I think we need to open a bug on the parse_rpm_name() method and also improve it so it works with a packagename which doesn't have the trailing ".rpm".  Possibly it should just be replaced with the code that implements _split_packagename().  But first we should check that it is actually causing problems where it is being used.

eherget:
This PR should be ready for merging.  I will do the following separately:

1) report bug in parse_rpm_name method mishandling epoch
2) write pkgtree unit tests under sprint task VMAAS-471 (increase unit tests to 80%)
3) investigate why repos with modules don't have basearch in pkgtree output

PR #521, opened 2019-06-20
add symlink for coverage command

psegedy:


webapp/Dockerfile-qe
```diff
@@ -17,6 +17,8 @@ ADD /webapp/*.txt /webapp/
 
 RUN pip3 install --upgrade pip && \
     pip3 install -r /webapp/requirements_qe.txt
+    
+RUN ln -s /opt/rh/rh-python36/root/bin/coverage /usr/bin/coverage
```
Josca: @psegedy I think it would be better to add link in the same place like other two links are created. It's not necessary to create new docker image layer for it (new `RUN` section).

PR #522, opened 2019-06-20
correct omitting of /opt in Jenkinsfile

psegedy:
Now, when it is not executed using scl-enable.sh, --omit path should not be in quotes.


psegedy:
Jenkins test failures are not connected with this PR, coverage report is now back in Artifacts.

PR #541, opened 2019-07-11
fixed missing values in swagger CvesResponse

Josca:


jdobes:
I'm not sure if it's worth fixing since @tkasparek is working on a new openapi spec during connexion re-factoring?

PR #542, opened 2019-07-12
Add process to upgrade/check new module versions.

yungbender:
simillar to: [engine](https://github.com/RedHatInsights/vulnerability-engine/pull/378)


scripts/upgrade-requirements.sh
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/bash
```
Josca: I have the same comment as [here](https://github.com/RedHatInsights/vulnerability-engine/pull/378#discussion_r302907840). What is purpose of this script?

PR #544, opened 2019-07-15
fix module parsing

eherget:
This should fix VMAAS-626

### DIFF
--- a/reposcan/repodata/modules.py
+++ b/reposcan/repodata/modules.py
@@ -22,24 +22,27 @@ class ModuleMD:
                 stream = data['stream']
                 if not name in module_dict:
                     module_dict[name] = {}
-                module_dict[name][stream] = {}
-                module_dict[name][stream]['name'] = name
-                module_dict[name][stream]['arch'] = data.get('arch')
-                module_dict[name][stream]['version'] = int(data.get('version'))
-                module_dict[name][stream]['context'] = str(data.get('context'))
-                module_dict[name][stream]['stream'] = str(stream)
-                module_dict[name][stream]['profiles'] = data.get('profiles', [])
-                for profile_name in module_dict[name][stream]['profiles']:
-                    module_dict[name][stream]['profiles'][profile_name]['default_profile'] = False
+                if not stream in module_dict[name]:
+                    module_dict[name][stream] = []
+                new_stream = {}
+                new_stream['name'] = name
+                new_stream['arch'] = data.get('arch')
+                new_stream['version'] = int(data.get('version'))
+                new_stream['context'] = str(data.get('context'))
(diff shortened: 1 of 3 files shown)


eherget:
I pushed an update yesterday late (2019-07-16) when I saw some log messages reporting multiple modules for pkg_id, repo_id.  I had put this in originally (when doing the modules in pkgtree output) because at that time there should not have been more than one module for a pkg_id and repo_id combination.  With the recent fix to make sure module and stream is reported in pkgtree for different versions of the same package where appropriate, the code was changed in a way that there could be multiple modules for a pkg_id, repo_id combination.

The change to the code now results in multiple instances of a module dict, one per stream.  So if a module has more than one stream and the streams only differ by versions, there likely will be more than one module dict for a pkg_id, repo_id combination.

For example:
perl-hivex-1.3.15-6.module+el8.0.0+3075+09be6b65.x86_64
is in the 'virt' module.  This package is linked to 3 streams, all are "rhel" stream and context is "55190bc5" with 3 different versions:
8000020190516125745
8000020190529063309
8000020190510171727
Our code, internally, represents this as the package being linked to 3 module dicts, one for each of the different versions and all with module name 'virt' and stream name 'rhel'.  Only the module name and stream name are put into the pkgtree output (and only once, not 3 times).

reposcan/repodata/modules.py
```diff
@@ -22,39 +22,45 @@ def __init__(self, filename):
...
+            if default_stream and default_stream in module_dict[name]:
+                for stream_dict in module_dict[name][default_stream]:
+                    stream_dict['default_stream'] = True
             for stream in data['profiles']:
                 if stream not in module_dict[name]:
                     continue  # this is here due to module defaults referencing non-existing modules
                 default_profile = data['profiles'][stream]
-                if default_profile and default_profile[0] in module_dict[name][stream]['profiles']:
-                    module_dict[name][stream]['profiles'][default_profile[0]]['default_profile'] = True
+                if default_profile:
```
Josca: This section is not covered by tests. Is it possible to fix it? Maybe by adding some data to testing file `modules.yaml`. As you can see everything but this section is covered: <https://codecov.io/gh/RedHatInsights/vmaas/src/d0eb311780953ab2d6f6ddc287620d225aed0f98/reposcan/repodata/modules.py>

Josca: Otherwise I can merge it.

eherget: This should be fixed now - https://codecov.io/gh/RedHatInsights/vmaas/src/a26bb0a099b9b0428b8f9ba8d69715a0b6f60864/reposcan/repodata/modules.py

PR #545, opened 2019-07-16
vmaas support for SSL certificates passed via env variable in reposcan

Michael-Jurek:


reposcan/reposcan.py
```diff
@@ -285,13 +285,23 @@ def _parse_input_list(self):
...
+
         for repo_group in data:
             # Entitlement cert is optional
             if "entitlement_cert" in repo_group:
                 cert_name = repo_group["entitlement_cert"]["name"]
-                ca_cert = repo_group["entitlement_cert"]["ca_cert"]
-                cert = repo_group["entitlement_cert"]["cert"]
-                key = repo_group["entitlement_cert"]["key"]
+
+                ca_cert = os.getenv(repo_group["entitlement_cert"]["ca_cert"][1:], "string") if \
```
jdobes: just a detail - instead of "string" default value, please make sure it'll return variable name (if it's not found in environment)

`ca_cert = os.getenv(repo_group["entitlement_cert"]["ca_cert"][1:], repo_group["entitlement_cert"]["ca_cert"])`

same for cert and key

reposcan/reposcan.py
```diff
@@ -48,13 +51,15 @@ def __init__(self, msg, success=True):
...
     """Base handler setting CORS headers."""
+
     def data_received(self, chunk):
         pass
 
     def set_default_headers(self):
         self.set_header("Access-Control-Allow-Origin", "*")
         self.set_header("Access-Control-Allow-Headers", "Content-Type")
-        self.set_header("Access-Control-Allow-Methods", "GET, PUT, POST, DELETE")
+        self.set_header("Access-Control-Allow-Methods",
```
jdobes: please update the PR to not do such unnecessary line wraps

if we want to do a re-formatting, let's do it in separate task

Michael-Jurek: I am sorry about it, this had to be a job of `ctrl + shift + i`  formatting in vscode.

PR #548, opened 2019-07-16
Add pkgtree_change item in dbchange get API.

yungbender:


reposcan/pkgtree.py
```diff
@@ -89,6 +89,15 @@ def dump(self):
             LOGGER.info("Removing old dump %s", fname)
             os.unlink(fname)
 
+        self._update_pkgtree_timestamp()
+
+    def _update_pkgtree_timestamp(self):
```
tkasparek: Could you please insert here timestamp of the pkgtree (defined on line 58) instead of current timestamp, so the values returned by the /dbchange API and in /pkgtree output match?

eherget:
@tkasparek 
A couple things here about trying to put pkgtree export timestamp into the db to be served up by webapp...
The db export timestamp gets written to the dbdump by exporter when it creates the dump.  This results in the "dbchange:exported" value being available to webapp to serve up via its dbchange api, however the export timestamp is never written to the database.  Its only in the dump used by webapp.
Issue 1)  This implementation had pkgtree write the value to the database so that when exporter dumps the database, the value is there.  This would be a new thing to have pkgtree write to the database.  Before this, pkgtree only read from the database and exporter continues to only read from the database.  Since the pkgtree is not served up by webapp, I would argue it is not necessary for webapp to serve up when pkgtree was last updated.  Reposcan serves up pkgtree output file, so any api for its last run should only need to be served there.
Issue 2) If pkgtree is to write its last run timestamp to the database, then pkgtree must be run just before the dbdump, not just after as it currently is.

PR #549, opened 2019-07-17
/repos API + modified_since wrong and inconsistent results

blayson:
`curl -X POST -d '{"repository_list": ["rhel-7-server-rpms"]}' localhost:8080/api/v1/repos/ | python -m json.tool`

```{
    "page": 1,
    "page_size": 2,
    "pages": 1,
    "repository_list": {
        "rhel-7-server-rpms": [
            {
                "basearch": "x86_64",
                "label": "rhel-7-server-rpms",
                "name": "Red Hat Enterprise Linux 7 Server (RPMs)",
                "product": "Red Hat Enterprise Linux Server",
                "releasever": "7.4",
                "revision": "None",
                "url": "https://cdn.redhat.com/content/dist/rhel/server/7/7.4/x86_64/os/"
            },
            {
                "basearch": "x86_64",
                "label": "rhel-7-server-rpms",
                "name": "Red Hat Enterprise Linux 7 Server (RPMs)",
                "product": "Red Hat Enterprise Linux Server",
                "releasever": "7Server",
                "revision": "2018-04-10T09:02:26+00:00",
                "url": "http://vmaas-data-vmaas-qe.5a9f.insights-dev.openshiftapps.com/repodata-for-vmaas/rhel/server/7/7Server/x86_64/os/"
            }
        ]
    }
}
```
The problem was that API returns repository list which contain list of repositories by label key, so modified_since parameter doesn't work and does not filter any repo, also problem with page size has the same roots

### DIFF
--- a/webapp/repos.py
+++ b/webapp/repos.py
@@ -87,6 +87,7 @@ class RepoAPI:
                 repo_details[label] = self.cache.repo_detail[repo_id]
         filters.append((filter_item_if_exists, [repo_details]))
 
+        actual_page_size = 0
         repo_page_to_process, pagination_response = paginate(repos, page, page_size, filters=filters)
         for label in repo_page_to_process:
             for repo_id in self.cache.repolabel2ids.get(label, []):
@@ -101,10 +102,13 @@ class RepoAPI:
                     "product": repo_detail[REPO_PRODUCT],
                     "revision": repo_detail[REPO_REVISION]
                 })
+            actual_page_size += len(repolist[label])
 
         response = {
             'repository_list': repolist,
         }
+
+        pagination_response['page_size'] = actual_page_size
         response.update(pagination_response)
         if modified_since:
             response["modified_since"] = modified_since
--- a/webapp/repos.py
+++ b/webapp/repos.py
@@ -44,13 +44,20 @@ class RepoAPI:
 
         return [label for label in self.cache.repolabel2ids if re.match(repo_regex, label)]
 
+    @staticmethod
+    def _modified_since(repo_detail, modified_since_dt):
+        if not modified_since_dt or (repo_detail[REPO_REVISION] != 'None' and parse_datetime(
+                repo_detail[REPO_REVISION]) > modified_since_dt):
+            return True
+        return False
+
     def _filter_modified_since(self, repos_to_process, modified_since_dt):
+        """Filter repositories by modified since"""
         filtered_repos_to_process = []
         for label in repos_to_process:
             for repo_id in self.cache.repolabel2ids.get(label, []):
                 repo_detail = self.cache.repo_detail[repo_id]
-                if not modified_since_dt or (repo_detail[REPO_REVISION] != 'None' and parse_datetime(
-                        repo_detail[REPO_REVISION]) > modified_since_dt):
+                if not modified_since_dt or self._modified_since(repo_detail, modified_since_dt):
                     filtered_repos_to_process.append(label)
         return filtered_repos_to_process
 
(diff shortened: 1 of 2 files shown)


webapp/repos.py
```diff
@@ -87,24 +87,29 @@ def process_list(self, api_version, data):  # pylint: disable=unused-argument
...
-                repolist.setdefault(label, []).append({
-                    "label": label,
-                    "name": repo_detail[REPO_NAME],
-                    "url": repo_detail[REPO_URL],
-                    "basearch": none2empty(repo_detail[REPO_BASEARCH]),
-                    "releasever": none2empty(repo_detail[REPO_RELEASEVER]),
-                    "product": repo_detail[REPO_PRODUCT],
-                    "revision": repo_detail[REPO_REVISION]
-                })
+                if not modified_since_dt or (repo_detail[REPO_REVISION] != 'None' and parse_datetime(
```
Josca: I would rather write `if not modified_since_dt or self._modified_since(repo_detail, modified_since_dt): ...`.

Josca:
@blayson are you going to do something with that?

blayson:
Yep, sure :smile:

webapp/test/test_repos.py
```diff
@@ -1,14 +1,18 @@
...
 
 import pytest
+import pytz
 
 from repos import RepoAPI
 
 REPO_JSON_EMPTY = {}
 REPO_JSON_BAD = {"page_size": 9}
 REPO_JSON = {"repository_list": ["rhel-7-server-rpms"]}
+REPO_JSON_MODIFIED_SINCE = {"repository_list": ["rhel-7-server-rpms"], "modified_since": "2014-04-06T01:23:45+02:00"}
```
Josca: You set some `"modified_since"` value here but you then change it in the test. Why? Isn't it easier directly to set it to value you need for that test?

webapp/test/test_repos.py
```diff
@@ -63,3 +67,15 @@ def test_schema(self):
...
         assert schemas.repos_schema.validate(response)
+
+    def test_modified_since(self):
+        """Test repos API with 'modified_since' property."""
+        repo = REPO_JSON_MODIFIED_SINCE.copy()
+        repo["modified_since"] = str(datetime.datetime.now().replace(tzinfo=pytz.UTC))
+        response = self.repo.process_list(api_version=1, data=repo)
+        assert tools.match(EMPTY_RESPONSE, response) is True
+
+        # without modified date
```
Josca: Here are two tests in one test method. If tests are independent (and these seem to be) it's better to add them each to separate test method.

webapp/test/test_repos.py
```diff
@@ -63,3 +67,15 @@ def test_schema(self):
...
+        repo["repository_list"] = ["REPO-W/O-MODIFIED"]
```
Josca: I don't know what does this trick means. If I want to use request "without modified data", I use json without `"modified_since"` field. Why do you use setting `"repository_list"` field to this strange value?

webapp/test/test_repos.py
```diff
@@ -63,3 +67,15 @@ def test_schema(self):
...
+        response = self.repo.process_list(api_version=1, data=repo)
+        assert tools.match(EMPTY_RESPONSE, response) is True
```
Josca: Both cases test only negative (empty) response. But I think it's more important to test positive response when you get some data from server. You fixed some paging feature, so I would expect adding tests where this feature is used.

webapp/test/test_repos.py
blayson: It sets an actual date, so whenever tests would be run it always will be the actual date

Josca: Yes, but what is it good for? Why not simply set today timestamp there?

blayson: :thinking: maybe, you right, I just made in a similar way how it was in errata and cve tests

Josca: Yes, I see it. It seems to be originally created by @psegedy. Guys @tkasparek @jdobes @eherget, what do you think about it? I think it would be better to rewrite it everywhere to constant timestamp instead of datetime.now(). It's better to have deterministic tests.

tkasparek: +1 to that, but I'd fix this just for this PR and would fix the rest in another PR.

Josca: @tkasparek thanks, I agree.

webapp/test/test_repos.py
Josca: @tkasparek @jdobes do you guys know what does the trick with `"REPO-W/O-MODIFIED"` mean?

webapp/test/test_repos.py
blayson: Ok, I'll do it

webapp/test/test_repos.py
blayson: I removed it from repos, but in errata as I understand because of this
```python
# make errata_detail without ERRATA_UPDATED
errata_detail2 = self.cache.errata_detail[ERRATA_NAME]
errata_detail_list2 = list(errata_detail2)
errata_detail_list2[ERRATA_UPDATED] = None
self.cache.errata_detail["RHSA-W/O:MODIFIED"] = errata_detail_list2
```

we set especially to this **errata** `modified_date/updated` status to `None`, so after we can test specifically this errata with update status `None`

PR #554, opened 2019-07-26
Added deps versions checker

Josca:
- and updated deps

### DIFF
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -9,7 +9,8 @@ services:
       CI: "true"
       TRAVIS: "true"
     command: >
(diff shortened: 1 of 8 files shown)


jdobes:
Good but it's creating even more differences compared to https://github.com/RedHatInsights/vulnerability-engine/blob/master/run_tests.sh, do we need separate scripts/check_deps_versions.sh script? there is no main requirements.txt referencing other files etc.

Josca:
@jdobes I created separate script to enable easily usage as a separate command e.g. in dev workspace. Also I excluded it from run_tests.sh script to run it only once.

eherget:
This looks good.

PR #558, opened 2019-07-30
better NamedCursor error handling

eherget:
This updates error handling in code that uses NamedCursor, which is exporter.py and pkgtree.py.

### DIFF
--- a/reposcan/pkgtree.py
+++ b/reposcan/pkgtree.py
@@ -10,6 +10,7 @@ import json
 import gzip
 from common.logging import get_logger, init_logging
 from common.dateutil import format_datetime, now
+from common.fileutil import remove_file_if_exists
 from database.database_handler import DatabaseHandler, NamedCursor, init_db
 
 DEFAULT_KEEP_COPIES = "2"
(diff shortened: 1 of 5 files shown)


reposcan/pkgtree.py
```diff
@@ -59,35 +59,41 @@ def dump(self):
...
+        except Exception: # pylint: disable=broad-except
+            # database exceptions caught here
+            LOGGER.exception("Failed to export pkgtree")
+        else:
+            # only write pkgtree if all db queries succeeded
+            LOGGER.info("Exporting data to %s", dump_filename)
+            with gzip.open(dump_filename, 'wt') as dump_file:
+                json.dump(self.outputdata, dump_file, indent=self.pkgtree_indent, ensure_ascii=False)
+            # relink to the latest file
+            try:
```
Josca: This construct is used 3 times here. Isn't it better to create `remove_if_exists(filename)` method somewhere in utils? Also I would rather use `os.remove` method. It is identical but has more familiar name for me.

eherget: I have pushed a new commit to cover this request.  Possibly it should be in a separate PR as it touches more files than the original commit.

Obviously I'm from the C programming era where unlink was the function to remove a file.  ;-)
"unlink" actually describes what unix/linux does under the covers better than "remove" does.

PR #559, opened 2019-07-31
FIX /repos API + modified_since wrong and inconsistent results

blayson:


webapp/repos.py
```diff
@@ -57,6 +57,8 @@ def _filter_modified_since(self, repos_to_process, modified_since_dt):
         for label in repos_to_process:
             for repo_id in self.cache.repolabel2ids.get(label, []):
                 repo_detail = self.cache.repo_detail[repo_id]
+                if label in filtered_repos_to_process:
+                    continue
```
eherget: I think the 'if label in filtered_repos_to_process:' could be moved before the repo_detail assignment statement above... if we're going to skip this label, might as well avoid doing anything else.  :)

blayson: Sure :)

webapp/repos.py
```diff
@@ -56,6 +56,8 @@ def _filter_modified_since(self, repos_to_process, modified_since_dt):
         filtered_repos_to_process = []
         for label in repos_to_process:
             for repo_id in self.cache.repolabel2ids.get(label, []):
+                if label in filtered_repos_to_process:
```
jdobes: seems working but shouldn't this condition be rather in parent for cycle?

this way it's going to do many unnecessary loops

blayson: Yes, done

PR #560, opened 2019-08-05
errata endpoint provides info about module updates

Michael-Jurek:
"module_list": {
"N, S, V, C":  [pkg_names]
}


webapp/test/test_app.py
```diff
@@ -80,12 +80,16 @@ def test_errata_post_1(self):
         """Test errata post endpoint."""
         body = """{"errata_list": [".*"]}"""
         resp = self.fetch('/api/v1/errata', method='POST', body=body)
+        print(resp.code)
```
jdobes: please remove print lines

webapp/app.py
```diff
@@ -987,6 +987,16 @@ def setup_apispec(handlers):
                             "example": "classification-RHSA-2018:0512"
                         }
                     },
+                    "module_list": {
```
jdobes: it's called module_list but it's an object, should by type of "array" instead of "object" to be consistent with other fields

also it'd be better to call it module**s**_list like in other APIs

perhaps change it to following format?
```
"modules_list": [
    {
        "module_name": "postgresql",
        "module_stream": "10",
        "module_version": "820190104140132",
        "module_context": "9edba152",
        "package_list": []
    }
]
```

or

```
"modules_list": [
    {
        "name": "postgresql",
        "stream": "10",
        "version": "820190104140132",
        "context": "9edba152",
        "package_list": []
    }
]
```

because I don't think `postgresql:10:820190104140132:9edba152` is some standard format

reposcan/exporter.py
```diff
@@ -253,6 +253,25 @@ def _dump_errata(self, dump):
                         errataid2bzs.setdefault(errata_id, []).append(ref_name)
                     else:
                         errataid2refs.setdefault(errata_id, []).append(ref_name)
+
+            # Select module ID to errata ID and package ID mapping
+            errata2modules = {}
```
jdobes: this dict should be called more like module2pkgid

PR #562, opened 2019-08-06
Add few reposcan and webapp unit-tests.

yungbender:
I added few tests for the vmaas. exactly:
- tests for the batchlist data structure
- tests for splay tree implementation used in webapp (updates.py)
- tests for rpm parsing in reposcan

### DIFF
--- a/reposcan/test/test_batchlist.py
+++ b/reposcan/test/test_batchlist.py
@@ -1,0 +1,45 @@
+"""
+Tests for batch list module.
+"""
+import pytest
+from common.batch_list import BatchList
+
+
+class TestBatchList:
+    """ Batch list tests group. """
+    batch_list = None
+    DUMMY_ITEMS = [{"item": 1}, {"item": 2}, {"item": 3}]
+
+    @pytest.fixture(autouse=True)
+    def setup_batch_list(self):
+        """ Fixture for creating batch list. """
+        self.batch_list = BatchList()
+
+    def test_empty(self):
+        """ Test for empty batch list. """
+        assert self.batch_list.get_total_items() == 0
+
+    def test_insert_few(self):
+        """ Test for inserton of one item. """
+        for item in self.DUMMY_ITEMS:
+            self.batch_list.add_item(item)
+
+        assert self.batch_list.get_total_items() == len(self.DUMMY_ITEMS)
+
+    def test_insert_full_batch(self):
+        """ Test for insertion of multiple items, until new batch needs to be created. """
+        for count in range(0, 51):
+            self.batch_list.add_item({"item": count})
+
+        assert self.batch_list.get_total_items() == 51
+        assert len(self.batch_list.batches) == 2
+        assert len(self.batch_list.batches[0]) == 50
+        assert len(self.batch_list.batches[1]) == 1
+
+    def test_clear(self):
+        """ Test for clearing the list. """
+        for item in self.DUMMY_ITEMS:
+            self.batch_list.add_item(item)
+
+        self.batch_list.clear()
+        assert self.batch_list.batches == []
--- a/webapp/test/test_updates.py
+++ b/webapp/test/test_updates.py
@@ -1,9 +1,6 @@
 """Unit tests for updates module."""
 # pylint: disable=protected-access
 # pylint: disable=unused-argument
-
-import os
-
 from test import schemas
 from test.conftest import TestBase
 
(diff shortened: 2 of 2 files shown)


reposcan/test/test_batchlist.py
```diff
@@ -0,0 +1,54 @@
...
+Tests for batch list module.
+"""
+import pytest
+from common.batch_list import BatchList
+
+
+class TestBatchList:
+    """ Batch list tests group. """
+    batch_list = None
+    dummy_items = []
```
Josca: It seems you can simply set `dummy_items = [{"item": 1}, {"item": 2}, {"item": 3}]` and then remove `setup_dummy_items`.

Josca: Also it's convention to name constant vars with upper case.

reposcan/test/test_rpm.py
```diff
@@ -0,0 +1,35 @@
+"""
+Tests for rpm packages name parsing.
+"""
+import pytest
+import common.rpm as rpm
+
+
+
```
Josca: One empty line more against PEP8.

reposcan/test/test_rpm.py
```diff
@@ -0,0 +1,35 @@
...
+        name, epoch, version, release, arch = rpm.parse_rpm_name(self.test_rpms[0])
+        assert name == "fedora-release-workstation"
+        assert epoch is None
+        assert version == "30"
+        assert release == "1"
+        assert arch == "noarch"
+
+    def test_rpmname_parse2(self):
+        """ Test for parsing RPM package name number 2. """
+        name, epoch, version, release, arch = rpm.parse_rpm_name(self.test_rpms[1])
```
Josca: This method `rpm.parse_rpm_name` belongs to `common` package. It already contains tests for this method. So it seems this file is useless :disappointed: .

webapp/test/test_updates.py
```diff
@@ -169,3 +169,72 @@ def test_process_list_empty_json(self):
             # NOTE: use copy of dict with json input, because process_list changes this dict
             self.updates_api.process_list(1, UPDATES_JSON_EMPTY.copy())
         assert "'package_list' is a required property" in str(context)
+
+class TestSplayTree(TestBase):
```
Josca: On the top of file is the comment `# TODO: write unit tests for Hot Cache`. Isn't it prepared for these tests?

webapp/test/test_updates.py
```diff
@@ -169,3 +169,72 @@ def test_process_list_empty_json(self):
...
+    """ Set of tests for Splay tree implementation. """
+    hot_cache = HotCache()
+    test_nevra = []
+
+    @pytest.fixture(autouse=True)
+    def setup_dummy_data(self):
```
Josca: Again, it's easier to directly set `test_nevra` (better `TEST_NEVRA`) collection in class head than append it later in method.

webapp/test/test_updates.py
```diff
@@ -169,3 +169,72 @@ def test_process_list_empty_json(self):
...
+        self.hot_cache.insert(self.test_nevra[1]["key"], self.test_nevra[1]["content"])
+
+    @pytest.fixture
+    def setup_huge_tree(self):
+        """ Fixture creates huge tree for pruning tests, etc. """
+        for order in range(0, 1025):
+            self.test_nevra[0]["key"] = order
+            self.hot_cache.insert(self.test_nevra[0]["key"], self.test_nevra[0]["content"])
+
+
```
Josca: One empty line more.

webapp/test/test_updates.py
```diff
@@ -169,3 +169,72 @@ def test_process_list_empty_json(self):
...
+    @pytest.fixture
+    def setup_small_tree(self):
+        """ Fixture creates small tree for testing. """
+        self.hot_cache.insert(self.test_nevra[2]["key"], self.test_nevra[2]["content"])
+        self.hot_cache.insert(self.test_nevra[0]["key"], self.test_nevra[0]["content"])
+        self.hot_cache.insert(self.test_nevra[1]["key"], self.test_nevra[1]["content"])
+
+    @pytest.fixture
+    def setup_huge_tree(self):
```
Josca: This design requires to insert method name as an argument. But it doesn't return any value. In this case it's better to create simple helper method `_setup_huge_tree` and call it at the start of touched tests.

webapp/test/test_updates.py
```diff
@@ -169,3 +169,72 @@ def test_process_list_empty_json(self):
...
+        """ Test for counting inserts. """
+        assert self.hot_cache.inserts == 3
+
+    def test_update_key(self, setup_small_tree):
+        """ Test for updating node content in tree. """
+        updated_content = {"reponse": "updatednevra2"}
+
+        self.hot_cache.insert(self.test_nevra[2]["key"], updated_content)
+
+        test_node = self.hot_cache.find(self.test_nevra[2]["key"])
```
Josca: As I see the tests, maybe it would be better to split your testing collection into `test_nevra_keys` and `test_nevra_contents`. It avoids permanent writing of `["keys"]` and `["content"]`.

Josca: Ok, I see you don' like this my proposal :cry: .

Josca:
@yungbender As you removed rpm_parse tests, you should also update PR description.

webapp/test/test_updates.py
yungbender: No, its not like that, but I just think that having the content and key together is more self explanatory for next code reader :). Also saving the few characters in line is not that ground breaking. (just my opinion)

yungbender:
Good point.

PR #563, opened 2019-08-07
webapp_utils added to devel docker-compose, changed entrypoint, updat…

Michael-Jurek:
…ed docs


doc/developer_mode.md
```diff
@@ -23,23 +23,32 @@ Now you have bash inside vmaas_webapp container, run:
 
 ```./entrypoint.sh```
 
-3. Switch into **websocket** container:
+3. Switch to **webapp_utils** container:
+~~~bash
+docker-compose exec vmaas_webapp_utils bash
+~~~
+
```
Josca: Empty line more.

webapp_utils/entrypoint.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+DIR=$(dirname $0)
+
+exec $DIR/webapp_utils.py
```
Josca: This produces log message:
~~~
WARNING: This is a development server. Do not use it in a production deployment.
vmaas-webapp-utils
Use a production WSGI server instead.
~~~
You should use production server here (e.g. gunicorn). See vulnerability engine - manager. It's the same case.

webapp_utils/webapp_utils.py
```diff
@@ -30,4 +31,4 @@ def set_default_headers(response): # pylint: disable=unused-variable
 application = create_app() # pylint: disable=invalid-name
 
 if __name__ == "__main__":
-    application.run(port=8083)
+    application.run(host='0.0.0.0', port=8083)
```
Josca: I see also, that there is `health` function in utilities.py module. 
1. If it is handler called through GET /v1/health, it should be called from app. Now it's not.
2. I would prefer module utils.py rather than utilities.py, as it's simpler. But it's more common to store several frequently used and handy functions in such module, not web handler method.

PR #568, opened 2019-08-27
Refactoring of common utilities for vmaas parts

yungbender:


yungbender:
The biggest problem is that the pipenv file needs to be created for whole vmaas project folder, this pipfile needs to have every package, with the moving of common .py files into common directory in the root pylint cannot find the import modules, so I needed to edit the ```run_test.sh``` file to run the pylint from root directory, but pipenv for the root directory did not exist so that is why there is new pipfile in root.

PR #587, opened 2019-09-19
Migrate webapp to connexion

semtexzv:
This PR migrates webapp from Tornado to Connexion and aiohttp.
These changes incur some performance penalties in the worst case, since current implementation is 
performing further input validation, even after connexion validated input body.

Some quick testing showed that latencies under load have worsened, but # Req/s remains roughly the same, and latency under normal conditions is unaffected. [Results](https://drive.google.com/open?id=1ewR-4tpAoCqp7ROyTLyjFmp33-6oh6MOgaCPVvfP9Y0) ( Results from this commit are in the aiohttp columns)

### DIFF
--- a/webapp/Dockerfile-qe
+++ b/webapp/Dockerfile-qe
@@ -30,5 +30,6 @@ EXPOSE 8080
 ADD /webapp/*.spec.yaml /webapp/
 ADD /webapp/*.sh        /webapp/
 ADD /webapp/*.py        /webapp/
+ADD /common/*.py        /webapp/common/
 
 CMD /webapp/entrypoint-qe.sh
--- a/webapp/Dockerfile-qe
+++ b/webapp/Dockerfile-qe
@@ -10,6 +10,8 @@ RUN yum -y install centos-release-scl && \
 
 ENV PATH=/opt/rh/rh-python36/root/usr/bin:$PATH
 
+WORKDIR /webapp
+
 ADD /webapp/Pipfile* /webapp/
 
 ENV LC_ALL=en_US.utf8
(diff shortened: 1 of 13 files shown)


webapp/Dockerfile-qe
```diff
@@ -26,8 +26,9 @@ ENV COVERAGE_FILE='/tmp/.coverage'
 
 EXPOSE 8080
 
-ADD /webapp/*.sh     /webapp/
-ADD /webapp/*.py     /webapp/
-ADD /common/*.py     /webapp/common/
```
psegedy: please add back `ADD /common/*.py     /webapp/common/`
Jenkins is failing because of
```
Everything is up - executing command
  File "webapp/app.py", line 18, in <module>
    from cache import Cache
  File "/webapp/cache.py", line 9, in <module>
    from common.logging_utils import get_logger
ModuleNotFoundError: No module named 'common'
```

semtexzv: Fixed, must have been merge error.

psegedy:
Looks like something isn't right because it failed with
```
Step 4/4 Wait for webapp to load latest data and refresh cache

script returned exit code 1
```
that is just `curl -s "${webapp_url}/api/v1/updates/bash-4.2.45-5.el7_0.4.x86_64`, so webapp is not responding
in openshift logs I don't see any tracebacks, nor messages that webapp is connected to websocket or cache is refreshed

semtexzv:
How does the webapp testing script launch the application ? The main module used to be `app.py` but was changed to `main.py`

psegedy:
This is the issue here, thanks for investigating. It;s launched by app.py https://github.com/RedHatInsights/vmaas/blob/master/Jenkinsfile#L122
Can you change all occurrences of `webapp/app.py` in [Jenkinsfile](https://github.com/RedHatInsights/vmaas/blob/master/Jenkinsfile) to correct value, please?

PR #598, opened 2019-10-04
NIST CVEs in vulnerability engine bug fix in vmaas

blayson:
This bug fix related to task with this PR RedHatInsights/vulnerability-engine#427

Add deleting CVEs from DB if CVE was removed from cvemap.xml


reposcan/database/cvemap_store.py
```diff
@@ -86,23 +86,33 @@ def _update_cves(self, to_update):
...
+        cur.execute("""select id, name, source_id,
                                  description, impact_id,
                                  published_date, modified_date,
                                  cvss3_score, cvss3_metrics,
                                  iava, redhat_url,
                                  secondary_url, source_id,
-                                 cvss2_score, cvss2_metrics
-                                 from cve
-                                 join (values %s) t(name)
-                                using (name)""", cve_names, page_size=len(cve_names))
```
jdobes: removing rhis code causes an error when CVE is removed from CVE map and then re-occurs again later:

> vmaas-reposcan        | f82db87bb331 2019-11-25 16:40:46,356 database.cve_common: [ERROR] Failure while importing CVEs'Traceback (most recent call last):\n  File "/reposcan/database/cvemap_store.py", line 48, in _import_cves\n    list(to_import), page_size=len(to_import))\n  File "/opt/rh/rh-python36/root/usr/lib64/python3.6/site-packages/psycopg2/extras.py", line 1281, in execute_values\n    cur.execute(b\'\'.join(parts))\npsycopg2.errors.UniqueViolation: duplicate key value violates unique constraint "cve_name_key"\nDETAIL:  Key (name)=(CVE-1999-0018) already exists.\n'|

reposcan/database/cvemap_store.py
```diff
@@ -44,7 +44,10 @@ def _import_cves(self, to_import, cve_data):
                                                         cvss3_score, cvss3_metrics, iava,
                                                         redhat_url, secondary_url, source_id,
                                                         cvss2_score, cvss2_metrics)
-                                              values %s returning id, name""",
+                                              values %s on conflict on constraint cve_name_key
+                                              do update set source_id = (select id from cve_source
```
jdobes: this doesn't update any other column than source_id, when for example description changes in meanwhile, you need to sync twice to get it correctly synced, please make sure everything is updated on first sync, rather not copy update code here but re-use `_update_cves` method

PR #599, opened 2019-10-04
Fix bug 500 in webap-utils /errata

Michael-Jurek:


webapp_utils/errata.py
```diff
@@ -167,9 +167,12 @@ def _errata_exists(self, errata):
...
         db_connection = self.db_pool.get_connection()
         with db_connection.get_cursor() as cursor:
-            cursor.execute("""select name from errata
-                              where name = '%s'""" % errata)
-            errata_detail = cursor.fetchone()
+            try:
+                cursor.execute("""select name from errata
+                                  where name = '%s'""" % errata)
+                errata_detail = cursor.fetchone()
+            except Exception as _: # pylint: disable=broad-except
```
tkasparek: I think you could use just `except Exception` without naming it if I am not mistaken.
Also, could more specific exception, more related to the issue be used so we don't catch there everything (this can be problematic in the future, as here we would silently consume all the exceptions without logging it).

webapp_utils/errata.py
```diff
@@ -156,9 +158,13 @@ def _errata_exists(self, errata):
         """Checks if given errata is in db."""
         db_connection = self.db_pool.get_connection()
         with db_connection.get_cursor() as cursor:
-            cursor.execute("""select name from errata
-                              where name = '%s'""" % errata)
-            errata_detail = cursor.fetchone()
+            try:
+                cursor.execute("""select name from errata
+                                  where name = '%s'""" % errata)
```
jdobes: http://initd.org/psycopg/docs/usage.html#the-problem-with-the-query-parameters

"Warning Never, never, NEVER use Python string concatenation (+) or string parameters interpolation (%) to pass variables to a SQL query string. Not even at gunpoint." :-)

Michael-Jurek: :cold_sweat:

webapp_utils/errata.py
```diff
@@ -142,23 +148,25 @@ def _fill_errata(self, search):
...
-                    errata_list.append(item[0])
+            errata = "%" + search + "%"
+            cursor.execute("""select name from errata
+                              where name like %s""", (errata,))
+            query = cursor.fetchall()
+            for item in query:
+                errata_list.append(item[0])
         return errata_list
 
     def _errata_exists(self, errata):
```
semtexzv: This will cause us to run N+1 queries for N erratas. In the end, this method is redundant, since process_list in `errata.py` performs a query, which returns all erratas.

Michael-Jurek: Fixed by [PR 613](https://github.com/RedHatInsights/vmaas/pull/613)

PR #606, opened 2019-10-14
Allow downloading repolist from git

semtexzv:
This PR adds a new SyncTask, which downloads repolist from github and imports it into database.

Before merging we have to figure out whether we want to create a new Openshift secret , or use existing `github-vulnerability-bot` secret for authorization against private repository.


jdobes:
added some fixes to finish this feature and also things missed by #620

PR #608, opened 2019-10-14
Delete apidoc container and its remains

yungbender:
- Delete apidoc and its remains (openshift deployment, docker-compose, jenkins)
- Fix dependencies so webapp-utils can have its own swagger UI
All swagger UI should run on ```locahost:port/api/ui```. (swagger UI made from connexion)


webapp_utils/app.py
```diff
@@ -12,12 +12,13 @@
 
 def create_app():
     """ Creates the application for webapp utils. """
-    utils_app = connexion.App("webapp-utils", options={'swagger_ui': False,
-                                                       'openapi_spec_path': 'openapi.json'})
+    utils_app = connexion.App("webapp-utils", options={'swagger_ui': True,
+                                                       'openapi_spec_path': '/v1/apispec'})
```
jdobes: just keep it `openapi.json` and also change webapp and reposcan to have `openapi.json` as well

there is no need to have `/v1/apispec` anymore

yungbender:
> and this test is now failing, you can decide if it's make sense to change it and check openapi.json URL or to delete it because it's functionality built in connexion to provide this API and perhaps it's pointless to test it
> 
> test/test_app.py::TestWebappSupportMethods::test_apispec[pyloop] FAILED

I think you are right and the test should be deleted

PR #609, opened 2019-10-15
Memory optimizations

semtexzv:
- Store list[int] as raw long arrays, reduces memory usage by ~300 MB for production database.
- Deduplicate strings from package details and also store package_details as array , saves additional ~200 MB.


jdobes:
I did some basic testing - sync repos, get package details, get updates, seems working

PR #614, opened 2019-10-24
Fix memory issues caused by new updates encoding

semtexzv:
Updates from different but compatible architectures are no longer encoded in the dump file, instead, the code loops through all compatible architectures for each checked package.


psegedy:
ClientWebSocketResponse.send_str was never awaited and bunch of ISEs, see Artifacts in Jenkins job - vmaas-webapp pod logs
```
app: [INFO] Connected to: ws://vmaas-websocket.vmaas-qe.svc:8082/
cache: [INFO] Loaded data version 2019-10-24T11:21:55.077729+00:00.
app: [INFO] Cached data refreshed.
/webapp/app.py:337: RuntimeWarning: coroutine 'ClientWebSocketResponse.send_str' was never awaited
  self.websocket.send_str(msg)
app: [ERROR] Internal server error <8727179098444>: please include this error id in bug report.'Traceback (most recent call last):\n  File "/webapp/app.py", line 71, in handle_request\n    res = api_endpoint.process_list(api_version, data)\n  File "/webapp/updates.py", line 442, in process_l
...
```

semtexzv:
I simply cannot preserve the old behavior with new encoding, therefore closing this, and will be rolling back the previous changes which resulted in increased memory usage.

PR #622, opened 2019-11-12
Podman developer mode

Michael-Jurek:
Fix for starting app in developer mode with podman. More Info in Readme.md.
Example:
# Build containers
```
sudo ./scripts/devel-compose up --build
```
# Switch into container
```
sudo podman exec -it vmaas-webapp bash
```

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -14,7 +14,7 @@ testing-postgresql = ">=1.3.0"
 [packages]
 flask = "==1.1.1"
 aiohttp = "==3.6.0"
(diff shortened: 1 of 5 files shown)


Pipfile
```diff
@@ -26,6 +26,7 @@ tornado = ">=6.0.3"
 pyyaml = ">=5.1.2"
 requests = ">=2.22.0"
 podman-compose = {file = "https://github.com/containers/podman-compose/archive/devel.tar.gz"}
+podman = ">=1.6.0"
```
MichaelMraka: 1.4.4 should be enough (version released on RHEL7)

PR #636, opened 2019-12-02
support to return all updates

jdobes:
- export all errata to dump
- add new /v3/updates API to return all updates/security only if `security_only` parameter is set


psegedy:
/v1/errata is now returning all errata, not only errata with CVE. This is functional change in v1 api, but I understand it may be difficult to keep the old behavior. Do we know of someone who is using /v1/errata, and this change could eventually break their workflow?

jdobes:
I think it doesn't matter that /v1/errata now knows about more errata than before, user just requests details about errata list, he can for example request to get details about all RHSA-2019-* using regexp

PR #640, opened 2019-12-05
deleting vmaas repos fails due to module foreign key references

blayson:


blayson:
> what if module is refrenced from module_stream?
> 
> https://redhatinsights.github.io/vulnerability-docs/vmaas_database_master/relationships.html

do we need to implement cascade deleting or will it be better to make nullable reference in module or module stream?

jdobes:
let's delete it from all connected tables, I think it doesn't make sense to keep it with nullable ref

but do not add ON DELETE CASCADE constraints, do deleting manually in this function only

reposcan/database/repository_store.py
```diff
@@ -131,12 +131,24 @@ def delete_content_set(self, content_set_label):
             cur.execute("""select id from repo where content_set_id = %s""", (content_set_id,))
             repo_ids = cur.fetchall()
             for repo_id in repo_ids:
+                cur.execute("select id from module where repo_id = %s", (repo_id,))
+                module_id = cur.fetchone()
```
jdobes: check the cardinality, I think there can be many modules per repo

blayson: :thinking: sure, I missed it :smile:

PR #642, opened 2019-12-16
feat(database): add database upgrade scripts and tutorial

yungbender:
Basically, the process is the same like for vulnerability-engine.

- The database upgrade scripts are placed into ```/database/upgrade_scripts/``` folder
- Sources for the database upgrade process are in ```/database/upgrade``` folder
- Upgrade script is started from the reposcan container and is placed in root folder of the container (```dbupgrade.sh```)
- Also I wrote some tutorial in ```README.md``` inside database folder

### DIFF
--- a/Dockerfile.test
+++ b/Dockerfile.test
@@ -17,6 +17,7 @@ ADD /reposcan/Pipfile*     /vmaas/reposcan/
 ADD /websocket/Pipfile*    /vmaas/websocket/
 ADD /webapp_utils/Pipfile* /vmaas/webapp_utils/
 ADD Pipfile*               /vmaas/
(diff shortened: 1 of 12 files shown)


yungbender:
> Sorry for very late review. The upgrade test you've added is a fine unit test but doesn't have essential functionality from https://github.com/RedHatInsights/vulnerability-engine/blob/master/tests/database_tests/test_upgrade.py - comparing that 2 databases - first database created from latest vmaas_db_postgresql.sql and second database created from version 0 vmaas_db_postgresql.sql + applied all available upgrade scripts - are exactly same

I understand, I thought this unit test is supposed to test the script, not the upgrade scripts and correctness in the upgrade script directory. I will create a new one.

PR #644, opened 2020-01-03
alert when repo/CVE import in reposcan fails

blayson:
add metrics to count failed imports of cves and repos

### DIFF
--- a/reposcan/redhatcve/cvemap_controller.py
+++ b/reposcan/redhatcve/cvemap_controller.py
@@ -16,6 +16,7 @@ from redhatcve.cvemap import CvemapHead, CvemapBody
 URL = os.getenv('REDHAT_CVEMAP_URL',
                 'https://www.redhat.com/security/data/metrics/cvemap.xml')
 
+
 class CvemapController:
     """
     Controls import/sync of CVE map into the DB.
@@ -36,8 +37,7 @@ class CvemapController:
 
     def _download_head(self):
         item = DownloadItem(source_url=URL,
-                            target_path=self._tmp_head()
-                           )
+                            target_path=self._tmp_head())
         download_items = [item]
         self.downloader.add(item)
         self.downloader.run(headers_only=True)
@@ -52,7 +52,7 @@ class CvemapController:
 
             # already synced before?
             db_lastmodified = parse_datetime(self.cvemap_store.lastmodified())
-            #db_lastmodified = None
+            # db_lastmodified = None
             self.lastmodified = parse_datetime(header.get_lastmodified())
             # synced for the first time or has newer revision
             if (db_lastmodified is None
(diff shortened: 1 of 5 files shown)


reposcan/redhatcve/cvemap_controller.py
```diff
@@ -97,5 +97,7 @@ def store(self):
                 self._download_xml()
                 cvemap = self._load_xml(self.lastmodified)
                 self.cvemap_store.store(cvemap)
+        except Exception:  # pylint: disable=broad-except
```
jdobes: probably not neccessary here because errors are catched inside self.cvemap_store.store already

reposcan/repodata/repository_controller.py
```diff
@@ -224,6 +225,7 @@ def import_repositories(self):
                 failures += 1
         if failures > 0:
             self.logger.warning("Failed to import %d repositories.", failures)
+            FAILED_IMPORT_REPO.inc(failures)
```
jdobes: increase counter also in `def store(self):` method, this handles only repo import but not repo sync

reposcan/reposcan.py
```diff
@@ -519,6 +519,7 @@ def run_task(*args, **kwargs):
         except Exception as err:  # pylint: disable=broad-except
             msg = "Internal server error <%s>" % err.__hash__()
             LOGGER.exception(msg)
+            FAILED_IMPORT_CVE.inc()
```
jdobes: increase also FAILED_IMPORT_REPO in this file

reposcan/redhatcve/cvemap_controller.py
blayson: :thinking: ok

reposcan/repodata/repository_controller.py
```diff
@@ -293,6 +295,8 @@ def md_size(repomd, data_type):
                                          completed_repositories, total_repositories)
                         self.repo_store.store(repository)
                         repository.unload_metadata()
+                except Exception:  # pylint: disable=broad-except
+                    FAILED_IMPORT_REPO.inc()
```
jdobes: but this will silently hide the exception

nevermind, for repositories I think it will be enough to only increase counters on top level in reposcan.py

blayson: wouldn't it be better to raise an exception right after `FAILED_IMPORT_REPO.inc()`?

jdobes: I'd not do that, it'd increment FAILED_IMPORT_REPO twice (second time on top level in reposcan.py), I think it's unnecessary

reposcan/repodata/repository_controller.py
```diff
@@ -224,8 +225,9 @@ def import_repositories(self):
                 failures += 1
         if failures > 0:
             self.logger.warning("Failed to import %d repositories.", failures)
+            FAILED_IMPORT_REPO.inc(failures)
 
-    def store(self):
+    def store(self):  # pylint: disable=too-many-branches,too-many-statements
```
jdobes: just remove this pylint comment

PR #650, opened 2020-01-14
Add pkgtree API endpoint to webapp

matysek:
The goal of this pull request to add new API endpoint 'pkgtree' that will return list of NEVRAs for given package name. The returned data format is similar to the 'pkgtree' API from reposcan app.

Two benefits of this API:
- no authentication required
- return NEVRA list just for one package name
- will be usable for insights-core


psegedy:
Sorry for not working Jenkins. I run integration tests locally and tests passed.

Out of curiosity, why did you pin all packages to `==` instead of `>=` and use exploitable `unsafe_load` instead of `full_load`? `Werkzeug` is not needed IMO

matysek:
@psegedy 
* I added a new dependency on package ```natsort``` for sorting response items. I added this then to ```Pipfile``` and wanted to regenerate ```Pipfile.lock```. However, I couldn't find a way how to update ```Pipfile.lock``` with just one package. Everytime all packages in Pipfile.lock were updated.
* Package `pyyaml v5.3` caused failures with webapp tests because with pyyaml 5.3 was changed ```full_load``` to not support python tags and just ```unsafe_load``` supported that and allowed to load ```test/data/cache.yaml```. I was getting this error:
```
self = <test.yaml_cache.YamlCache object at 0x7f4ae19aa1d0>    def load_yaml(self):
        """Load cache from YAML file."""
        with open(self.filename, "r") as stream:
            try:
                data = yaml.full_load(stream)
            except yaml.YAMLError as err:
                print(err)>       for key, val in data.items():
E       UnboundLocalError: local variable 'data' referenced before assignment
could not determine a constructor for the tag 'tag:yaml.org,2002:python/object/apply:decimal.Decimal'
  in "test/data/cache.yml", line 25, column 5
```
* Werkzeug was installed automatically as a dependency of something. Also with updated version of ```Werkzeug``` I got another test failure related to this:
```
DeprecationWarning: The import 'werkzeug.FileStorage' is deprecated and will be removed in Werkzeug 1.0. Use 'from werkzeug.datastructures import FileStorage' instead.
    from werkzeug import FileStorage
```
* So I fiddled with various combinations of package versions and the only way how to have a stabilized environment when updating ```Pipfile.lock``` was by setting package versions in ```Pipfile``` to ```==```.
* When not using pyyaml 5.3 method ```unsafe_load``` could be changed back to ```full_load```.
* When ```Werkzeug``` is not needed how to prevent it from being installed? Or is it only a dev dependency of something?
* Do you know about any way how to update ```Pipfile.lock``` with just ```natsort``` and avoid updating all other packages?

psegedy:
@matysek thanks for explanation. I'm not familiar with updates of `Pipfile`. It just looked strange to me why another WSGI is required for the application, but it might be needed for unit-tests... I'm sure developers know better.

PR #659, opened 2020-02-17
return null errata severity instead of string "None"

blayson:
fix(database): store null severity in errata table
fix(reposcan): import null severity to db
fix(webapp): return null instead of empty string

should I create 3 different commits or 1 will be ok?


jdobes:
1 commit is fine

please add also nullable flag in openapi spec

blayson:
> 1 commit is fine
> 
> please add also nullable flag in openapi spec

Done

PR #661, opened 2020-02-18
Incorrect response with incorrect content type

blayson:


webapp/app.py
```diff
@@ -401,42 +401,41 @@ def create_app():
...
     @web.middleware
-    async def error_formater(request, handler, **kwargs):
-        #pylint: disable=broad-except
+    async def error_formatter(request, handler, **kwargs):
+        # pylint: disable=broad-except
         def build_error(detail, status):
             errors = {"detail": detail, "status": status}
             return {"errors": [errors]}
 
+        original_error = ''
```
jdobes: what is the purpose of this variable? it's a string type and few lines below this variable changes to dict

blayson: it was added by mistake :smile:

PR #670, opened 2020-02-24
filter errata by type, severity etc.

blayson:


webapp/webapp.spec.yaml
```diff
@@ -741,6 +741,23 @@ components:
...
+            - type: string
+              example: 'security'
+            - type: array
+              items:
+                type: string
+                example: security
+        severity:
+          oneOf:
+            - type: string
+              example: 'Critical'
```
jdobes: shouldn't this be nullable as well?

blayson: hmm, I fixed it
but I have tested it before and it was working fine 🤷‍♂

jdobes: are you testing it with upgraded schema (null severity_id columns in errata table)? and current webapp dump (generated by reposcan)?

jdobes:
any update?

PR #671, opened 2020-02-26
chore: upgrade connexion version

yungbender:
The old connexion version is causing a problem when we need to upgrade packages through Pipfile.
[Source](https://github.com/zalando/connexion/issues/1149)


jdobes:
there is one test failure - different error message for empty input, please check if it's related to this change and fix the test if needed

yungbender:
> there is one test failure - different error message for empty input, please check if it's related to this change and fix the test if needed

Yes, I'm working on it, looks like they added their own error formatter for aiohttp.

PR #689, opened 2020-04-08
don't block APIs when doing refresh and return 503 instead

jdobes:
- healthchecks won't fail because webapp stays responsive
- other apps can detect refreshing state by 503 code and retry instead of time out


jdobes:
tests failing because of different issue

PR #690, opened 2020-04-14
chore: disable pipenv check because of pyup.io API key issues

tkasparek:


MichaelMraka:
I think it will be better to have it enabled and create an option (ARG/ENV) to be able to disable it in the local and test deployment

jdobes:
closing in favor of #691

PR #692, opened 2020-04-15
feat(webapp): new API to list only applicable errata to a package list

MichaelMraka:
new /patches API to show only simple errata list


jdobes:
rebased on master to fix travis ci test

PR #693, opened 2020-04-16
chore: allow to get PIPENV_CHECK variable from host environment

jdobes:
allows commands like:
$ PIPENV_CHECK=0 docker-compose up --build


matysek:
* @jdobes could you document this variable in the README.md?
* @jdobes what about putting ```PIPENV_CHECK=0``` to script file ```./scripts/devel-compose```

jdobes:
@matysek added mention to readme and devel-compose has it turned off by default now

PR #702, opened 2020-04-22
Fix rpms and srpms responses

yungbender:
- Now the package will be displayed as empty ("package_name" {}) only when it exists in the database.


blayson:
Update branch to master please, there was fixed test_upgrade

PR #709, opened 2020-05-15
Added gzip response middleware

Josca:
- Added option `GZIP_RESPONSE_ENABLE` to gzip response.
- Implemented transparently using middleware.
- Compress level is configurable using `GZIP_COMPRESS_LEVEL` env var, 5 by default.


conf/webapp.env
```diff
@@ -4,3 +4,4 @@ WEBSOCKET_HOST=vmaas_websocket
 MAX_VMAAS_SERVERS=1
 LOGGING_LEVEL=INFO
 LOGGING_TYPE=DEVEL
+GZIP_RESPONSE_ENABLE=on
```
jdobes, line 7: in code is default value "FALSE", there is "on", please make both consistent "TRUE"/"FALSE" or "on"/"off"

webapp/app.py
```diff
@@ -536,11 +552,15 @@ def format_error(detail, status):
 
         return res
 
+    middlewares = []
+    if GZIP_RESPONSE_ENABLE:
+        middlewares.append(gzip_middleware)
+    middlewares.extend([error_handler, timing_middleware])
```
jdobes, line 558: can you initialize on line 555?

```middlewares = [error_handler, timing_middleware]```

or does order of middlewares matter?

File: webapp/app.py [hunk 3]
```diff
@@ -517,6 +521,18 @@ async def timing_middleware(request, handler, **kwargs):
 
         return res
 
+    @web.middleware
+    async def gzip_middleware(request, handler, **kwargs):
+        """ Middleware that compress response using gzip"""
+        res = await handler(request, **kwargs)
+        LOGGER.info(request.headers)
```
jdobes, line 528: make it LOGGER.debug please, info is default level here and logs would be overly verbose

conf/webapp.env
Josca: ok, also added level param to .env file

webapp/app.py
Josca: unfortunately, order matters

webapp/app.py
Josca: oh, sorry, that's forgotten log, removing it

PR #713, opened 2020-05-22
prevent unnoticed vmaas CDN cert expiration

blayson:


reposcan/repodata/repository_controller.py
```diff
@@ -76,24 +78,32 @@ def _download_repomds(self):
...
-                CERT_EXPIRATION.labels(cert_name).state('expired')
-            else:
-                self.logger.warning('Certificate not provided or incorrect')
-                CERT_EXPIRATION.labels('None').state('expired')
+            self.logger.warning('Certificate not provided or incorrect: %s', cert_name if cert_name else 'None')
+            msg = self._prepare_msg_for_slack(cert_name, 'Reposcan CDN certificate not provided or incorrect')
+            send_slack_notification(format_message, msg)
+
+    @staticmethod
+    def _prepare_msg_for_slack(cert_name, header, expire_tuple=None):
```
jdobes, line 98: I would probably move this function to `common/slack_notifications.py`.

reposcan/repodata/repository_controller.py
```diff
@@ -76,24 +78,32 @@ def _download_repomds(self):
...
-            if expire_in.days > 7:
-                self.logger.debug('Certificate %s is ok', cert_name)
-                CERT_EXPIRATION.labels(cert_name).state('valid')
-            elif 7 >= expire_in.days > 0:
-                self.logger.warning('Certificate %s will expire in %s', cert_name, expire_in.days)
-                CERT_EXPIRATION.labels(cert_name).state('expire_soon')
+            valid_to_dt = datetime.strptime(loaded_cert.get_notAfter(), "%Y%m%d%H%M%SZ")
+            expire_in_days_td = (valid_to_dt - datetime.utcnow()).days
+            expire_tuple = (valid_to_dt, expire_in_days_td)
+            if 7 >= expire_in_days_td > 0:
```
jdobes, line 84: Let's send the warning earlier, let's say 30 days. 7 days may not be enough time to issue new cert.

common/slack_notifications.py
```diff
@@ -10,8 +10,40 @@
...
+                        f":hourglass_flowing_sand: *Expire in*: {date['expire_in']} days"
+            }
+        })
+
+    return formatted
+
+
+def send_slack_notification(format_callback, msg: dict):
     """If SLACK_WEBHOOK is set, sends notification to Slack"""
+    formatted = format_callback(msg)
```
jdobes, line 46: Is there a reason why it's called as callback? Is the callback ever different method than format_message? If not I would call the format_message method directly so we don't have to import the method in repository_controller and pass it as an argument.

jdobes:
sorry for late review

common/slack_notifications.py
blayson: I was thinking in case we need different formatting in future for another notification

PR #714, opened 2020-05-25
package_names/srpms/<name> times out

blayson:
In python 3 method items() returns viewitems object, to use it effectively I added iter() function over it.

Currently request `package_names/srpms/kernel` takes about 100 secs on my local machine instead of return times out, which is a bit better 😄


blayson:
I got it, I created this PR in case they need a working solution asap. But I still working on it. I convert it to draft for now.

PR #716, opened 2020-05-26
fix(webapp): removed redundant product check

MichaelMraka:
including updates from different repos of the same product
confuses users because they see updates from repos which they do not
have enabled


MichaelMraka:
fixed

jdobes:
rebased on master to fix travis CI build

PR #723, opened 2020-06-11
feat(webapp): unify errata severity none to null

yungbender:
- The None severities are now changed to the SQL null

### DIFF
--- a/reposcan/database/test/conftest.py
+++ b/reposcan/database/test/conftest.py
@@ -66,18 +66,14 @@ def create_pg(vmaas_user=VMAAS_USER, vmaas_pg=VMAAS_PG):
             cursor.execute("GRANT ALL ON SCHEMA public TO public")
 
         with conn.cursor() as cursor:
-            pg_env = postgresql.dsn()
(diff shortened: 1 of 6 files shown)


reposcan/database/upgrade.py
```diff
@@ -36,6 +36,8 @@ class DatabaseUpgrade:
     def __init__(self):
         LOGGER.info('DatabaseUpgrade initializing.')
 
+        DatabaseHandler.close_connection()
```
yungbender, line 39: Here it is closing connection already in constructor, if there is some remaining connection from previous stuff. This made problems in tests.

reposcan/database/test/conftest.py
```diff
@@ -65,19 +65,15 @@ def _handler(postgresql):
             cursor.execute("GRANT ALL ON SCHEMA public TO postgres")
             cursor.execute("GRANT ALL ON SCHEMA public TO public")
 
-        with conn.cursor() as cursor:
```
yungbender, line 68: The custom user for tests was useless, there is one admin made automatically.

PR #728, opened 2020-06-23
Fixed webapp response gzipping

Josca:


webapp/app.py
```diff
@@ -527,16 +527,16 @@ async def gzip_middleware(request, handler, **kwargs):
...
-        if not request.path.startswith('/api/v'):  # exception not to gzip swagger pages
+        try:
+            if res.body is not None and header in request.headers and "gzip" in request.headers[header]:
+                gzipped_body = gzip.compress(res.body, compresslevel=GZIP_COMPRESS_LEVEL)
+                res.body = gzipped_body
+                res.headers["Content-Encoding"] = "gzip"
+                return res
+        except (TypeError, AttributeError):
+            LOGGER.warning("unable to gzip response '%s'", request.path)
             return res
```
jdobes, line 538: Nothing is returned if condition on L531 is false. There are two returns in gzip_middleware function, I would keep one return at the end.

webapp/app.py
```diff
@@ -527,16 +527,16 @@ async def gzip_middleware(request, handler, **kwargs):
         res = await handler(request, **kwargs)
         header = 'Accept-Encoding'
 
-        if not request.path.startswith('/api/v'):  # exception not to gzip swagger pages
+        try:
+            if res.body is not None and header in request.headers and "gzip" in request.headers[header]:
```
jdobes, line 531: Nothing is returned if condition on L531 is false. There are two returns in gzip_middleware function, I would keep one return at the end.

Josca: fixed

jdobes:
updated commit message to contain semantic-release keyword

PR #750, opened 2020-09-11
add option to return updates for latest packages only

MichaelMraka:
Fixing bz1875706
Adds new "latest_only" option to input json.

### DIFF
--- a/webapp/updates.py
+++ b/webapp/updates.py
@@ -4,8 +4,8 @@ Module to handle /updates API calls.
 
 from cache import REPO_LABEL, REPO_BASEARCH, REPO_RELEASEVER, REPO_URL, PKG_SUMMARY_ID, PKG_DESC_ID, \
     ERRATA_CVE, ERRATA_TYPE
-from common.webapp_utils import join_packagename, none2empty
-from common.rpm import parse_rpm_name
+from common.webapp_utils import none2empty
+from common.rpm import parse_rpm_name, join_rpm_name
 
 SECURITY_ERRATA_TYPE = 'security'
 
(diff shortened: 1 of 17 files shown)


webapp/updates.py
```diff
@@ -49,12 +50,13 @@ def _process_repositories(self, data, response):
...
         """Parse input NEVRAs and filter out unknown (or without updates) package names."""
-        packages_to_process = data.get('package_list', None)
+        latest_only = data.get("latest_only", False)
+        packages_to_process = filter_package_list(data.get('package_list', None), latest_only)
         filtered_packages_to_process = {}
         if packages_to_process is not None:
             for pkg in packages_to_process:
                 response['update_list'][pkg] = {}
-                name, epoch, ver, rel, arch = split_packagename(pkg)
+                name, epoch, ver, rel, arch = parse_rpm_name(pkg, default_epoch='0')
```
jdobes, line 59: Assuming `latest_only` is True, `parse_rpm_name` is called twice for every pkg, first in `filter_package_list` function, then here. It'd be good if the previously parsed result was re-used.

jdobes (changes requested):
I plan to check what is performace hit of `filter_package_list` function yet, if any.

webapp/updates.py
MichaelMraka: I'm aware of it.
Unnecessary parsing of 1k package names takes extra 2.1ms.
Returning already parsed list would make the code complicated and required joining components back (L#56) which takes 1.9ms for 1k packages.
Given the above I've decided to keep more readable code.

jdobes: ok, fair enough, I've also haven't measured any serious performance hit

jdobes (approved):

PR #763, opened 2020-10-08
Updated response size

AlesKas:
Reduced response size by 20%.

### DIFF
--- a/webapp/updates.py
+++ b/webapp/updates.py
@@ -124,7 +124,8 @@ class UpdatesAPI:
                 desc_id = self.db_cache.package_details[current_nevra_pkg_id][PKG_DESC_ID]
                 response['update_list'][pkg]['description'] = self.db_cache.strings.get(desc_id, None)
 
-            response['update_list'][pkg]['available_updates'] = []
+            if api_version != 3:
+                response['update_list'][pkg]['available_updates'] = []
 
             # No updates found for given NEVRA
             last_version_pkg_id = self.db_cache.updates[name_id][-1]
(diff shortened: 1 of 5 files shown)


blayson (changes requested):

webapp/updates.py
```diff
@@ -166,13 +171,21 @@ def _process_updates(self, packages_to_process, api_version, available_repo_ids,
...
+                        if api_version != 3:
+                            response['update_list'][pkg]['available_updates'].append({
+                                'package': nevra,
+                                'erratum': errata_name,
+                                'repository': repo_details[REPO_LABEL],
+                                'basearch': none2empty(repo_details[REPO_BASEARCH]),
+                                'releasever': none2empty(repo_details[REPO_RELEASEVER])
+                            })
+                        else:
+                            response['update_list'][pkg][nevra]['available_updates'].append({
```
jdobes, line 183: `pkg` is nevra of currently installed package, `nevra` is nevra of updated package, layering it like that and then accessing using `'available_updates'` is not optimal.

I was thinking more about something more structured like:

```
response['update_list'][pkg]['available_updates'] =
[
  {
    'package': nevra
    'errata': [
      {
        "name": "RHSA-2021-0001",
        "repositories": [{"label": label, "basearch": basearch, "releasever": releasever},]
      },
    ]
  },
]
```

also now, I'd rather make it as v4 API and not change v3

sorry for a very late review :slightly_frowning_face:

jdobes:
Pls re-open when needed.

PR #787, opened 2021-03-08
chore: use turnpike to sync repolist

psegedy:


scripts/setup_db.sh
```diff
@@ -71,13 +72,13 @@ fi
 headers=(-H "Authorization: token $token" -H "Content-type: application/json")
```
jdobes, line 72: The authorization token is no longer needed.

jdobes (changes requested):

jdobes (approved):

PR #790, opened 2021-03-19
VMAAS-1326: add third_party flag to repo and errata db tables

Josca:


Josca:
Included to #791, closing.

PR #792, opened 2021-03-23
chore: refactor and mirror health endpoint for clowder

yungbender:
Healthz endpoint in webapp is added through aiohttp instance, because the connexion API does not allow having empty basepath for aiohttp.

### DIFF
--- a/webapp/app.py
+++ b/webapp/app.py
@@ -12,7 +12,6 @@ from distutils.util import strtobool
 from json import loads
 
 import connexion
-import yaml
 from aiohttp import ClientSession
 from aiohttp import hdrs
 from aiohttp import web
@@ -52,6 +51,9 @@ GZIP_COMPRESS_LEVEL = int(os.getenv("GZIP_COMPRESS_LEVEL", "5"))
 LATEST_DUMP_ENDPOINT = "/api/v1/latestdump"
 
 LOGGER = get_logger(__name__)
+
+DEFAULT_PATH_API = "/api"
+DEFAULT_PATH = "/api/vmaas"
 
 
 class BaseHandler:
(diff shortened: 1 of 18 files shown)


psegedy:
@yungbender please delete readiness and liveness probes from https://github.com/RedHatInsights/vmaas/blob/master/deploy/clowdapp.yaml

yungbender:
> @yungbender please delete readiness and liveness probes from https://github.com/RedHatInsights/vmaas/blob/master/deploy/clowdapp.yaml

Should be resolved now.

webapp/app.py
```diff
@@ -618,20 +601,16 @@ async def on_prepare(request, response):  # pylint: disable=unused-argument
 
     app.app.on_response_prepare.append(on_prepare)
     app.app.router.add_get("/metrics", metrics)
+    app.app.router.add_get("/healthz", HealthHandler.get)
```
jdobes, line 604: why not use `healthz.spec.yaml` in webapp but use in reposcan?

yungbender: I wanted to use healthz spec for both, the problem is that the AIOHTTP app in webapp does not allow having empty api prefix, because connexion is building the AIOHTTP app through `add_subapp` function.

```
  File "/usr/local/lib/python3.6/site-packages/connexion/apps/aiohttp_app.py", line 59, in add_api
    self.app.add_subapp(api.base_path, api.subapp)
  File "/usr/local/lib64/python3.6/site-packages/aiohttp/web_app.py", line 302, in add_subapp
    raise ValueError("Prefix cannot be empty")
ValueError: Prefix cannot be empty

During handling of the above exception, another exception occurred:

  File "main.py", line 21, in <module>
    "": "healthz.spec.yaml"})
  File "/git/app.py", line 612, in create_app
    arguments={'vmaas_version': VMAAS_VERSION})
  File "/usr/local/lib/python3.6/site-packages/connexion/apps/aiohttp_app.py", line 62, in add_api
    "aiohttp doesn't allow to set empty base_path ('/'), "
connexion.exceptions.ConnexionException: aiohttp doesn't allow to set empty base_path ('/'), use non-empty instead, e.g /api
```

jdobes: ok, nm

jdobes (approved):

PR #793, opened 2021-03-24
added testing free repolist configs (fedora, epel)

Josca:


psegedy:
@Josca what is the purpose of this PR, are you going to run sync manually with these repolists? Because, these repolists can't be used for our integration tests

@jdobes the `retest` didn't work since the PR is not rebased. (the issue is that jenkins creates 'merge commit' locally, tries to deploy merged code, but the local 'merge commit' does not exist in the remote/origin, so it fails with 404. The error message should be clearer, though)

jdobes:
#827 was merged. Adding locked version of EPEL 8 for tests. I'd close this PR, perhaps the locked version of Fedora repo can be added too if we want to test Fedora.

PR #796, opened 2021-03-25
updates.py refactor

Josca:
Refactored webapp/updates.py to be easier to read.

### DIFF
--- a/webapp/test/test_updates.py
+++ b/webapp/test/test_updates.py
@@ -60,45 +60,17 @@ class TestUpdatesAPI(TestBase):
         """Setup UpdatesAPI object."""
         self.updates_api = UpdatesAPI(self.cache)
 
-    def test_process_repos(self):
-        """Test repos processing."""
-        response = {}
-        repo_ids = self.updates_api._process_repositories(UPDATES_JSON, response)
-        assert len(repo_ids) == 1
-
-    def test_process_repos_repo(self):
-        """Test repos processing - filter out repository_list."""
-        response = {}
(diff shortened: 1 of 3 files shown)


webapp/test/test_updates.py
```diff
@@ -62,49 +62,62 @@ def setup_api(self, load_cache):
...
-        response = {}
-        repo_ids = self.updates_api._process_repositories(UPDATES_JSON, response)
-        assert len(repo_ids) == 1
+        repositories_list, repo_ids = self.updates_api._get_repository_list(UPDATES_JSON)
+        releasever, repo_ids = self.updates_api._get_releasever(UPDATES_JSON, repo_ids)
+        basearch, repo_ids = self.updates_api._get_basearch(UPDATES_JSON, repo_ids)
+        assert repositories_list == ['rhel-7-server-rpms']
+        assert releasever == '7Server'
+        assert basearch == 'x86_64'
+        assert repo_ids == {1}
```
MichaelMraka, line 71: it no more tests self.updates_api._process_repositories() but 3 different methods

Josca: Well, these methods depend on each other and relate to repository_list processing so I kept them together. Should I split it to separate tests or just rename current test?

Josca: Ok, I've created a separate tests for the new functions. It will be better.

MichaelMraka: that's what I mean; keep test for old method and create one for new

MichaelMraka (approved):

jdobes (approved):

PR #811, opened 2021-04-13
feat: add sqlite database format

semtexzv:
Extract sqlite dump generation code from the semtezv/next branch


jdobes:
Added new OVAL dicts in #832 , I'm going to add them to this PR as well.

jdobes:
fixed missing things, validated functionality (also with integration tests (#842)), removed old code

LGTM

PR #815, opened 2021-04-15
Add pagination to "pkgtree" endpoint

Josca:


webapp/app.py
```diff
@@ -324,6 +324,15 @@ async def get(cls, package_name=None, **kwargs):
         return await cls.handle_request(cls.pkgtree_api, 1, 'package_name_list', package_name, **kwargs)
 
 
+class PkgtreeHandlerV3Get(BaseHandler):
```
semtexzv, line 327: Do we really need new class ? Is this a backwards incompatible change ?

Josca: No, it adds new fields into the response due to pagination. It's better to modify as less as possible, so I would change API version 3 only.

jdobes: I'd keep it as a new class too. This API is used also by other parties and better keep older version compatible.

jdobes (approved):

PR #834, opened 2021-05-04
add more as_long_arr casts

jdobes:


jdobes:
Will be eventually overwritten by #811 but let's add it still.

jdobes:
Need to fix CI

PR #837, opened 2021-05-06
feat(reposcan): allow repolists to opt_out of default certificates

semtexzv:
In current state we don't allow repolists without any cert. Either one is provided, or default is used. This allows opting out of the default cert.

### DIFF
--- a/reposcan/reposcan.py
+++ b/reposcan/reposcan.py
@@ -209,7 +209,7 @@ class SyncHandler:
     def finish_task(cls, task_result):
         """Mark current task as finished."""
         if cls not in (PkgTreeHandler, RepoListHandler, GitRepoListHandler, CleanTmpHandler) \
-            and "ERROR" not in task_result:
+                and "ERROR" not in task_result:
             # Notify webapps to update their cache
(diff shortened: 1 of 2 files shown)


reposcan/reposcan.py
```diff
@@ -256,11 +256,16 @@ def parse_repolist_json(cls, data):
         seen = set()
         for repo_group in data:
             # Entitlement cert is optional, use default if not specified in input JSON
-            if "entitlement_cert" in repo_group:
+            if "entitlement_cert" in repo_group and isinstance(repo_group['entitlement_cert'], object):
```
jdobes, line 259: Are you sure?

```
>>> isinstance({}, object)
True
>>> isinstance(False, object)
True
```

shouldn't it be `dict`?
```
>>> isinstance({}, dict)
True
>>> isinstance(False, dict)
False
```

jdobes (changes requested):

reposcan/reposcan.py
semtexzv: You're right, fixed.

PR #838, opened 2021-05-06
feat(reposcan): accept lists for content set data

semtexzv:
Current approach results in issues when using multiple baseurls. This change adds multiple entries per one content set, and later pull request will remove support for multiple baseurls. (Result will be the same, but we will be able to choose what baseurl template should be applied to what releasevers within one content set).

### DIFF
--- a/reposcan/reposcan.py
+++ b/reposcan/reposcan.py
@@ -222,10 +222,16 @@ class RepolistImportHandler(SyncHandler):
 
     @classmethod
     def _content_set_to_repos(cls, content_set):
+        all_repos = []
+
+        # Unpack list of content sets if provided
+        if isinstance(content_set, list):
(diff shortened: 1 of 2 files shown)


reposcan/reposcan.py
```diff
@@ -222,10 +222,16 @@ class RepolistImportHandler(SyncHandler):
 
     @classmethod
     def _content_set_to_repos(cls, content_set):
+        all_repos = []
+
+        # Unpack list of content sets if provided
+        if isinstance(content_set, list):
+            for cset in content_set:
+                all_repos += cls._content_set_to_repos(cset)
```
Josca, line 230: I don't like += notation too much, but ok.

Josca (approved):

jdobes (approved):

PR #839, opened 2021-05-10
chore: Reorganize project files as a Python project

Josca:


psegedy (changes requested):
deploy fails - at first sight it looks like init container is crashing because of https://github.com/RedHatInsights/vmaas/blob/master/deploy/clowdapp.yaml#L125
Edit: if we want to make it work in CI and QA envs it will probably need appropriate changes in e2e-deploy repo

psegedy (approved):

jdobes (approved):

PR #844, opened 2021-05-17
fix(reposcan): fix path to wait script and replace it with python version

jdobes:


Josca (approved):
great, to have one script less :+1:

PR #847, opened 2021-05-21
VULN-1347 refactor(webapp): make responses more consistent

yungbender:
- Removed returning of some filters: `modified_since`, `published_since`
- Added `last_change` in all endpoints


blayson (commented):
Why we are removed `modified_since` and `published_since`  from CVE response?

yungbender:
> Why we are removed `modified_since` and `published_since` from CVE response?

These two are just input filters, in other APIs we are not returning input filters. Thus removing it for better consistency of results.

blayson (approved):

PR #849, opened 2021-05-31
WIP: feat: implement requires_reboot flag for advisories

semtexzv:


vmaas/reposcan/repodata/updateinfo.py
```diff
@@ -33,6 +33,8 @@ def __init__(self, filename):
                 update["type"] = elem.get("type")
                 update["id"] = text_strip(elem.find("id"))
                 update["title"] = text_strip(elem.find("title"))
+                # Is the false here good choice ? Shouldn't the default be true ?
+                update["reboot"] = False
```
jdobes, line 37: I think `False` should be default. But in SQL default is `true`, I'd set `false` there as well to be consistent.

vmaas/reposcan/repodata/updateinfo.py
```diff
@@ -63,6 +65,10 @@ def __init__(self, filename):
                     for collection in pkglist.findall("collection"):
                         module = collection.find("module")
                         for pkg in collection.findall("package"):
+
+                            if pkg.get("reboot_suggested") in ["True", "true"]:
```
jdobes, line 69: possibly use `from distutils.util import strtobool` function?

jdobes (changes requested):

vmaas/reposcan/repodata/updateinfo.py
semtexzv: Thanks, included

vmaas/reposcan/repodata/updateinfo.py
jdobes: I see, however it'll result into incorrect reboot_suggested flags for most of advisories. We should find a way how to set it to correct value for all existing advisories in DB. Something like marking all repos revision to timestamp 0, so the reboot flag will be updated for all advisories in all repos (but I'm afraid we'd need to implement mechanism to update changed advisory metadata)

And second thing, I checked some updateinfo.xml and there is `reboot_suggested` flag for both advisory and packages in the advisory. Here we go through packages in advisory but we could just take the flag of the advisory and don't care about individual packages.

semtexzv: What about something like this. Please check my updated `update_store.py` I first do an on-conflict update, and then lookup of values for returning. We do the same in patchman-engine on many places and it seems to be the simplest way for insertin/updating and getting all referneced rows.

psegedy:
There are tracebacks in sync, see `ci.int.devshift.net PR build` job
```
vmaas.reposcan.repodata.repository_controller: [INFO] Syncing repository: vmaas-test-rhel8-module, x86_64, 8 [12/13]
vmaas.reposcan.database.object_store: [ERROR] Failure inserting into errata.'Traceback (most recent call last):\n  File "/vmaas/vmaas/reposcan/database/update_store.py", line 118, in _populate_updates\n    execute_values(cur, """select id, name from errata where name in (%s)""", list(names))\n  File "/usr/local/lib64/python3.6/site-packages/psycopg2/extras.py", line 1292, in execute_values\n    cur.execute(b\'\'.join(parts))\npsycopg2.errors.UndefinedFunction: operator does not exist: text = record\nLINE 1: select id, name from errata where name in ((\'R\',\'H\',\'S\',\'A\',...\n                                               ^\nHINT:  No operator matches the given name and argument types. You might need to add explicit type casts.\n'|
vmaas.reposcan.database.object_store: [ERROR] Failure changing package associations.'Traceback (most recent call last):\n  File "/vmaas/vmaas/reposcan/database/update_store.py", line 196, in _associate_packages\n    update_to_packages)\n  File "/vmaas/vmaas/reposcan/database/update_store.py", line 21, in _get_associations_todo\n    update_id = update_map[update["id"]]\nKeyError: \'RHSA-2018:2020\''|
vmaas.reposcan.database.object_store: [ERROR] Failure updating errata_cve data.'Traceback (most recent call last):\n  File "/vmaas/vmaas/reposcan/database/update_store.py", line 294, in _associate_cves\n    update_id = update_map[update["id"]]\nKeyError: \'RHSA-2018:2020\''|
vmaas.reposcan.repodata.repository_controller: [INFO] Syncing repository: rhel-7-server-rpms, ppc64, 7.7 [13/13]
vmaas.reposcan.database.object_store: [ERROR] Failure inserting into errata.'Traceback (most recent call last):\n  File "/vmaas/vmaas/reposcan/database/update_store.py", line 119, in _populate_updates\n    for row in cur.fetchall():\npsycopg2.ProgrammingError: no results to fetch'|
```

vmaas/reposcan/database/update_store.py
```diff
@@ -92,40 +92,34 @@ def _populate_updates(self, updates):
...
-                for row in cur.fetchall():
+                               solution, requires_reboot) values %s
+                               on conflict (name) do update set requires_reboot = excluded.requires_reboot
+            """, errata)
+
+            names = set()
+            for update in updates:
+                names.add(update["id"])
+
+            cur.execute("""select id, name from errata""")
```
jdobes, line 117: We can't select all errata here. This will cause for exaple that all errata in DB will get associated with repo.

semtexzv: We load all IDs from database, but then filter out the rows based on names on client. I don't know why, but I could'nt get the `where name in %s` condition to work.

jdobes (approved):

PR #851, opened 2021-06-03
feat(reposcan): retry periodic cache dump later if it failed

yungbender:


jdobes (commented):
Code seems ok, my only concern is that it'll end up in an error loop caused by some error that can't be fixed by retry. In this error loop it'll download 20K of repository files, cve map and OVAL files every 5 minutes.

What if we attempt a retry only for errors where is a chance that repeated run will succeed - like the DB connection problem. It'd require extending the output of a task to contain more than just an "ERROR" but should be possible.

What do you think?

yungbender:
> Code seems ok, my only concern is that it'll end up in an error loop caused by some error that can't be fixed by retry. In this error loop it'll download 20K of repository files, cve map and OVAL files every 5 minutes.
> 
> What if we attempt a retry only for errors where is a chance that repeated run will succeed - like the DB connection problem. It'd require extending the output of a task to contain more than just an "ERROR" but should be possible.
> 
> What do you think?

I like the idea, but I don't see the 5 minute loop. I mean, the periodic callback calls the new method `PeriodicSync` class, which performs the first export, and if that fails, the IOLoop `call_later` 5 minut callback calls only `AllSyncHandler`, which does not have the retry functionality.

jdobes:
> I like the idea, but I don't see the 5 minute loop. I mean, the periodic callback calls the sync method through`PeriodicSync` class, which has the `finish_task` method, which performs the first export, and if that fails, the IOLoop `call_later` 5 minut callback calls only `AllSyncHandler` class method, which does not have the retry functionality. So it retries only 1 time.

I see it now, thanks.

`if isinstance(err, (DatabaseError, OperationalError)):` - OperationalError is subclass of DatabaseError, couldn't only DatabaseError be here?

yungbender:
> > I like the idea, but I don't see the 5 minute loop. I mean, the periodic callback calls the sync method through`PeriodicSync` class, which has the `finish_task` method, which performs the first export, and if that fails, the IOLoop `call_later` 5 minut callback calls only `AllSyncHandler` class method, which does not have the retry functionality. So it retries only 1 time.
> 
> I see it now, thanks.
> 
> `if isinstance(err, (DatabaseError, OperationalError)):` - OperationalError is subclass of DatabaseError, couldn't only DatabaseError be here?

Oh, I thought they are different. Fixed.

PR #859, opened 2021-06-28
Fix multiple content-set syncing process

semtexzv:
Syncing a repolist with multiple items per single content_set resulted in an error. Fixing this.


jdobes (changes requested):
seems good, pylint only complains about trailing whitespace

jdobes (approved):

PR #888, opened 2021-09-03
VMAAS-1377: respect module requires

MichaelMraka:


jdobes (changes requested):
`module_stream_require` table is referencing `module_stream` table, references from this table need to be removed when calling reposcan API for deleting content sets

see delete_content_set function in vmaas/reposcan/database/repository_store.py

MichaelMraka:
stream requires removal fixed

jdobes (approved):

PR #898, opened 2021-09-16
fix(webapp): load as many tables as possible even when dump is not up to date

jdobes:


Josca (changes requested):
conflicts ...

vmaas/webapp/cache.py
```diff
@@ -150,58 +150,68 @@ def download():
         """Download new version of data."""
         return not os.system("rsync -a --copy-links --quiet %s %s" % (REMOTE_DUMP, DUMP))
 
+    @staticmethod
+    def _sqlite_execute(data, query):
+        try:
+            return data.execute(query)
+        except sqlite3.Error as err:
+            # file does not exist or has wrong type
+            LOGGER.warning("Failed to load data %s: %s", query, err)
```
Josca, line 159: Shouldn't we use "error" log level here?

Josca (approved):
Lgmt

PR #901, opened 2021-09-17
feat(webapp): enable optimistic updates for /vulnerabilities

jdobes:
VMAAS-1386

### DIFF
--- a/vmaas/webapp/vulnerabilities.py
+++ b/vmaas/webapp/vulnerabilities.py
@@ -151,6 +151,7 @@ class VulnerabilitiesAPI:
 
     def process_list(self, api_version, data):  # pylint: disable=unused-argument
         """Return list of potential security issues"""
+        data["optimistic_updates"] = True  # find updates even if original package is not found in repo
         evaluate_oval = data.get("oval", False)
         evaluate_oval_only = data.get("oval_only", False)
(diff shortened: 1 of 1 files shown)


vmaas/webapp/vulnerabilities.py
```diff
@@ -151,6 +151,7 @@ def _content_sets_to_definitions(self, content_set_list):
 
     def process_list(self, api_version, data):  # pylint: disable=unused-argument
         """Return list of potential security issues"""
+        data["optimistic_updates"] = True  # find updates even if original package is not found in repo
```
Josca, line 154: This way you will change existing API behaviour. Why don't you turn it on in request?

jdobes: `/vulnerabilities` API is calling `/updates` in the background. I wanted to hardcode it for `/vulnerabilities` because I didn't want to make it optional here. This API is combining results from `/updates` and OVAL (OVAL finds CVEs exactly this "optimistic" way), if `/updates` would be called with `optimistic_updates = False`, the combined results would be weird - some CVEs marked as manually fixable while they can be fixed by a playbook.

Josca: @jdobes ok then.

PR #913, opened 2021-11-04
chore: Add separated memory limits for services in clowdapp.yaml

Josca:


Josca:
for now resolved by #914

PR #916, opened 2021-11-04
fix(webapp): rebuild index array on cache reload

Josca:
- VMAAS-1391

### DIFF
--- a/vmaas/webapp/cache.py
+++ b/vmaas/webapp/cache.py
@@ -89,7 +89,6 @@ class Cache:
         self.filename = filename
         self.clear()
         self.reload()
-        self.build_indexes()
 
     def clear(self):
         """Clear dictionaries and load new data."""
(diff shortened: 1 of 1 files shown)


vmaas/webapp/cache.py
```diff
@@ -149,6 +149,7 @@ def reload(self):
         if self.download():
             self.clear()
             self.load_sqlite(self.filename)
+            self.build_indexes()
```
MichaelMraka, line 152: Now there's redundant `self.build_indexes()` in `__init__()` (L:92) .

MichaelMraka (changes requested):

vmaas/webapp/cache.py
Josca: Ah, thanks, fixed.

MichaelMraka (approved):

jdobes (approved):

PR #921, opened 2021-11-08
chore: VMAAS-1353 modify /packages to retun package name always

mkholjuraev:
After the PR:

image

### DIFF
--- a/vmaas/webapp/packages.py
+++ b/vmaas/webapp/packages.py
@@ -42,8 +42,14 @@ class PackagesAPI:
         want_third_party = data.get('third_party', False)
 
         packagelist = {}
+
+        response = {
+            'last_change': utils.format_datetime(self.cache.dbchange['last_change'])
+        }
+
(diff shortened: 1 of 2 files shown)


vmaas/webapp/packages.py
```diff
@@ -78,11 +78,13 @@ def process_list(self, api_version, data):  # pylint: disable=unused-argument
 
             # If the package is third party, then remove it from result
             if not want_third_party and is_third_party:
-                del packagelist[pkg]
+                packagelist[pkg] = {}
+
+        if len(packagelist) == 1 and list(packagelist.values())[0] == {}:
```
jdobes, line 83: `last_change` shouldn't be removed. There should be no difference in response when single package or more packages are requested.

jdobes (changes requested):

jdobes (approved):
failing db_change test is unrelated, LGTM

PR #930, opened 2021-11-29
chore: refresh singleton in wait_for_services

psegedy:


wait_for_services.py
```diff
@@ -30,8 +31,13 @@ def wait(func, *args, delay=1, service="", **kwargs):
                 return
             LOGGER.info("%s is unavailable - sleeping", service)
             time.sleep(delay)
-        except:  # noqa pylint: disable=bare-except
```
jdobes, line 33: pls keep also the broad except, if some other exception happens

jdobes (changes requested):

jdobes (approved):

PR #956, opened 2022-02-16
Update README for tests

michalslomczynski:
These are steps I used in order to run selected tests interactively with stdout from tests attached. If we want to keep our README concise and prefer to not have additional dockerfiles then maybe these changes are not necessary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,7 +118,23 @@ Congratulations!
 ### Run tests
 You can run all tests from scratch just after cloning repo using command:
 ~~~bash
-sudo podman-compose -f docker-compose.test.yml up --build --abort-on-container-exit
+sudo docker-compose -f docker-compose.test.yml up --build --abort-on-container-exit
+~~~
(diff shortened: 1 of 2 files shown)


Josca (commented):
I think this is just one of many approaches how to test this project. E.g. I prefer tests running from IDE. Sometimes I just temporarily modify dockre-compose.test.yml config to run some specific tests subset from it. I would not add another docker-compose file as we already have many of them here.

Also I wouldn't change podman-compose in README to docker-compose. We prefer podman to be used and every developer knows he can use docker as a container runtime if he wants.

@MichaelMraka @jdobes please write your opinion too.

MichaelMraka:
I'd like to stick with podman-compose as well. If something is not working in podman but it is in docker let's debug and fix it.

I agree we already have quite number of docker-compose*.yml files. What about modifying compose or image to accept (sub)set of tests through env variable and otherwise running all of them?

jdobes:
I agree, let's not add more docker-compose files.

PR #959, opened 2022-02-21
VMAAS-1411: Handle non existing key exception

michalslomczynski:


michalslomczynski:
API response after changes:

        {
          "type": "about:blank",
          "detail": "ValueError('Invalid arch name: x86_aaaa64',)",
          "status": 400
        }

vmaas/webapp/vulnerabilities.py
```diff
@@ -49,7 +49,10 @@ def _evaluate_state(self, state, epoch, ver, rel, arch):
 
         candidate_arches = self.db_cache.ovalstate_id2arches.get(oval_state_id, [])
         if candidate_arches:
-            matched = matched and self.db_cache.arch2id[arch] in candidate_arches
+            if self.db_cache.arch2id.get(arch) is not None:
```
tkasparek, line 52: I would say that more Python way for this condition is:
```if arch in self.db_cache.arch2id```

tkasparek (changes requested):

vmaas/webapp/vulnerabilities.py
michalslomczynski: Thanks for this suggestion. Did not know about it.

jdobes (changes requested):
Please change commit message to contain semantic release keywords, for example in this case:

fix(webapp): VMAAS-1411: Handle non existing key exception

jdobes:
Ok, next time :smile:

PR #1047, opened 2023-09-25
Fix security vulnerabilities.

shannon-donahue:
## Overview:
- Updating packages in order to fix security vulnerabilities.

### Testing instructions:
1. Pull branch and build docker image with new updated packages
2. Run the following to perform security scan: `docker run --rm --volume /var/run/docker.sock:/var/run/docker.sock --name Grype anchore/grype:latest "$(docker images --format '{{.Repository}}' |grep {insert image tag name})" --only-fixed`
3. The output should show no high or critical vulnerabilities.

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -18,96 +18,96 @@
     "default": {
         "aiohttp": {
             "hashes": [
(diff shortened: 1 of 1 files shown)


aleccohan (approved):
LGTM!

PR #1050, opened 2023-10-18
feat(webapp): return and display latest change of repo in vmaas

yungbender:
RHINENG-2608

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/vmaas/webapp/repos.py
+++ b/vmaas/webapp/repos.py
@@ -11,6 +11,7 @@ from vmaas.webapp.cache import (
     REPO_RELEASEVER,
     REPO_PRODUCT,
     REPO_REVISION,
+    REPO_LAST_CHANGE,
     REPO_THIRD_PARTY
 )
 from vmaas.common.date_utils import format_datetime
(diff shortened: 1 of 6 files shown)


vmaas/webapp/repos.py
```diff
@@ -105,6 +106,7 @@ def process_list(self, api_version, data):  # pylint: disable=unused-argument
 
         actual_page_size = 0
         repo_page_to_process, pagination_response = paginate(repos, page, page_size, filters=filters)
+        latest_repo_change = parse_datetime(self.cache.dbchange['last_change'])
```
jdobes, line 109: I think initializating to this value doesn't make much sense because `dbchange['last_change']` is the latest timestamp of all other values in dbchange. And it'll always be newer than any repo last_change.

jdobes: Let's initialize it to a first repo last_change for example. And the value can be null probably for a case when there is no repository in the output list.

PR #1063, opened 2024-01-11
feat(reposcan): add CSAF sync task

michalslomczynski:

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management


psegedy:
Tests are failing because sync now takes too long and also the local development with docker-compose will be harder because you'd need to wait at least an hour for the sync.
I'd suggest to do something similar as we are doing with OVAL and repolists. Let's add modified changes.csv (e.g. only with 2023 cves) and jsons referenced in changes.csv to https://github.com/RedHatInsights/vmaas-data. Then we can set env to point to vmaas-data repo and download files from there. The env would need to be set in reposcan.env and in app-interface vmaas ephemeral deploy target.
@mtclinton what do you think?
@michalslomczynski would you mind doing that?

michalslomczynski:
Will do, thanks for proposing resolution.

mtclinton:
yes, I like this solution :)

jdobes (approved):

PR #1066, opened 2024-01-16
Migration to Connexion 3 - attempt #2

jdobes:

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management


vmaas/webapp/app.py
```diff
@@ -434,7 +412,17 @@ class RPMPkgNamesHandlerGet(BaseHandler):
...
+
+
+def metrics():
+    """Provide current prometheus metrics"""
+    return generate_latest(), 200, {'Content-Type': 'text/plain; charset=utf-8'}
+
+
+def dummy_200():
+    """Dummy endpoint returning 200"""
+    return "", 200
```
psegedy, line 425: looks like a forgotten testing function

vmaas/webapp/app.py
```diff
@@ -503,105 +489,102 @@ def load_cache_to_apis():
...
+        async def __call__(self, scope, receive, send):
+            if scope["type"] != "http":
+                await self.app(scope, receive, send)
+                return
+
+            async def send_with_formatted_error(message):
+                if message["type"] == "http.response.start":
+                    self.start_message = message
+                elif message["type"] == "http.response.body":
+                    if self.start_message["status"] >= 400:
```
psegedy, line 529: is `self.start_message` set at this point, wouldn't it be an exception? I see it is set to `None` in `__init__` and then it is set only here in `__call__` but  not in this branch.

vmaas/webapp/app.py
jdobes: No, it's used for OPTIONS /v3/cves endpoint. Not sure if it's still used by something, I've just converted it.

vmaas/webapp/app.py
jdobes: The assumption here is that for a single response the middleware is called twice. First the `http.response.start` sections which handles headers and then the `http.response.body` section which handles body. It shouldn't happen that `self.start_message` is not set in the `http.response.body` section.

psegedy (approved):

PR #1073, opened 2024-02-12
feat(reposcan): Add CsafStore products insert

michalslomczynski:

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management


psegedy (changes requested):
it fails on foreign key constraint, please take a look
```
2024-02-13 14:28:16,053:ERROR:vmaas.reposcan.database.object_store:Failure while inserting, updating or deleting data to csaf_products: update or delete on table "csaf_products" violates foreign key constraint "csaf_product_id" on table "csaf_cves"DETAIL:  Key (id)=(17) is still referenced from table "csaf_cves".'Traceback (most recent call last):\n  File "/vmaas/vmaas/reposcan/database/csaf_store.py", line 149, in _populate_data\n    cur.execute(f"""delete from {table_name} where id in %s\npsycopg2.errors.ForeignKeyViolation: update or delete on table "csaf_products" violates foreign key constraint "csaf_product_id" on table "csaf_cves"\nDETAIL:  Key (id)=(17) is still referenced from table "csaf_cves".\n'|
2024-02-13 14:28:16,063:ERROR:vmaas.reposcan.database.object_store:Failure while inserting, updating or deleting data to csaf_cves: insert or update on table "csaf_cves" violates foreign key constraint "csaf_product_id"DETAIL:  Key (csaf_product_id)=(5965) is not present in table "csaf_products".'Traceback (most recent call last):\n  File "/vmaas/vmaas/reposcan/database/csaf_store.py", line 136, in _populate_data\n    execute_values(cur, f"""insert into {table_name} ({\', \'.join(cols)}) values %s\n  File "/usr/local/lib64/python3.11/site-packages/psycopg2/extras.py", line 1299, in execute_values\n    cur.execute(b\'\'.join(parts))\npsycopg2.errors.ForeignKeyViolation: insert or update on table "csaf_cves" violates foreign key constraint "csaf_product_id"\nDETAIL:  Key (csaf_product_id)=(5965) is not present in table "csaf_products".\n'|
```

vmaas/reposcan/database/csaf_store.py
```diff
@@ -46,7 +48,129 @@ def _save_csaf_files(self, csaf_files: CsafFiles) -> list[int]:
         self.conn.commit()
         return db_ids
 
+    def _get_product_rows(self, csaf_cves):
```
psegedy, line 51: I understand that you want to have the same signature for `_get_product_rows` and `_get_cve_rows` so you're returning but I think it would be better to use OOP for this, create a base class (as an interface in different OOP languages) with `get_rows` method raises just NotImplementedError and implement that method in classes inheriting from base class. It would improve readability and mainly maintainability.

vmaas/reposcan/database/csaf_store.py
```diff
@@ -46,7 +48,129 @@ def _save_csaf_files(self, csaf_files: CsafFiles) -> list[int]:
...
+        to_insert = []
+        to_update = None
```
psegedy, line 53: remove it from here since it is never used and just explicitly return `None` in return statement as `return to_insert, None`

vmaas/reposcan/database/csaf_store.py
```diff
@@ -46,7 +48,129 @@ def _save_csaf_files(self, csaf_files: CsafFiles) -> list[int]:
...
+        for _, csaf_products in csaf_cves.items():
+            for csaf_product in csaf_products:
+                product_values = csaf_product.to_tuple()
+                if not self.csaf_product_map.get(product_values):
+                    to_insert.append(product_values)
+                    # Insert just a key to the map to avoid duplicates.
+                    # Product ID will be updated on refresh cache step after row is inserted into DB.
+                    self.csaf_product_map[product_values] = -1
```
psegedy, line 61: do we need inserted values in this cache map? It requires to load all data from DB twice with this implementation and I don't think we really need to load all data from DB to python map

vmaas/reposcan/database/csaf_store.py
```diff
@@ -46,7 +48,129 @@ def _save_csaf_files(self, csaf_files: CsafFiles) -> list[int]:
...
+    def _product_refresh_cache(self, cur, delete=False):
+        for product_id, cpe, package, module in cur.fetchall():
+            if delete:
+                del self.csaf_product_map[(cpe, package, module)]
+            else:
+                self.csaf_product_map[(cpe, package, module)] = product_id
+
+    def _get_cve_rows(self, csaf_cves):
+        # TODO: Known_affected are only parsed as of now. Update here accordingly.
+        product_status_id = 4
```
psegedy, line 83: make this configurable with env variable, avoid magic constants in code if possible

vmaas/reposcan/database/csaf_store.py
```diff
@@ -46,7 +48,129 @@ def _save_csaf_files(self, csaf_files: CsafFiles) -> list[int]:
...
+    def _populate_cves(self, csaf_cves):
+        self.csaf_cve_map = self._prepare_table_map(cols=["cve", "csaf_product_id"],
+                                                    to_cols=["id", "product_status_id"],
+                                                    table="csaf_cves")
+
+        self._populate_data(csaf_cves, "csaf_cves", ["cve", "csaf_product_id", "product_status_id"],
+                            self._get_cve_rows, self._cve_refresh_cache,
+                            "(%s::int, %s, %s::int, %s::int)")
+
+    def _populate_data(self, csaf_cves, table_name, cols, get_rows, refresh_cache, update_tmpl_str):
```
psegedy, line 124: this can be done with OOP and inheritance to improve readability as well

vmaas/reposcan/database/csaf_store.py
```diff
@@ -46,7 +48,129 @@ def _save_csaf_files(self, csaf_files: CsafFiles) -> list[int]:
...
+                               f"""update {table_name} set {', '.join([f'{col} = v.{col}' for col in cols])}
+                                   from (values %s) as v(id, {', '.join(cols)})
+                                   where {table_name}.id = v.id
+                                   returning {table_name}.id, {', '.join([f'{table_name}.{col}' for col in cols])}
+                                """, tuple(to_update,), page_size=len(to_update), template=update_tmpl_str)
+                refresh_cache(cur)
+            self.conn.commit()
+        except Exception as exc:
+            CSAF_FAILED_IMPORT.inc()
+            CSAF_FAILED_UPDATE.inc()
```
psegedy, line 145: catch exceptions just after db query, don't update both IMPORT and UPDATE counters if you don't know if it failed with insert or update

vmaas/reposcan/database/csaf_store.py
```diff
@@ -46,7 +48,129 @@ def _save_csaf_files(self, csaf_files: CsafFiles) -> list[int]:
...
+                            "(%s::int, %s, %s::int, %s::int)")
+
+    def _populate_data(self, csaf_cves, table_name, cols, get_rows, refresh_cache, update_tmpl_str):
+        """Generic method to populate table with CSAF data."""
+        to_insert, to_update = get_rows(csaf_cves)
+        try:
+            cur = self.conn.cursor()
+            if to_insert:
+                execute_values(cur, f"""insert into {table_name} ({', '.join(cols)}) values %s
+                                        returning id, {', '.join(cols)}""",
```
psegedy, line 131: returning statement is never used, you are not fetching any values from `execute_values` and not storing it to any variable, please remove `returning ...` if you are not using it

vmaas/reposcan/database/csaf_store.py
```diff
@@ -46,7 +48,129 @@ def _save_csaf_files(self, csaf_files: CsafFiles) -> list[int]:
...
+                               to_insert, page_size=len(to_insert))
```
psegedy, line 132: `page_size=len(to_insert)` this does not seem to have any effect since you are setting the page_size to total number of inserted items. It can be removed I think

vmaas/reposcan/redhatcsaf/modeling.py
```diff
@@ -167,6 +171,10 @@ def update(self, data: CsafCves) -> None:
         """Update data in collection - same as dict.update()."""
         self._cves.update(data._cves)  # pylint: disable=protected-access
 
+    def items(self):
```
psegedy, line 174: please add unit tests for this and type hint

vmaas/reposcan/redhatcsaf/modeling.py
```diff
@@ -127,6 +127,10 @@ class CsafProduct:
     package: str
     module: Optional[str] = None
 
+    def to_tuple(self):
```
psegedy, line 130: please add unit tests and type hing

psegedy (changes requested):
In addition to specific points in the code I think the to_insert and to_update logic can be changed completely. Currently you are loading all data from DB into a map, then insert values and update values. After all that you are again loading all data from DB to map to refresh cache.
What I think you can try instead is to use a select query to find items which already exist in DB and update them. Then you can insert the rest and you won't need any `csaf_product_map ` and most importantly you won't need to load all data from DB to the map.

Other than that, please use type annotations so mypy can find issues with the code and IDE can show you what each functions expects and returns. Use specific types, don't just use something like `def fn(csaf_cves: dict) -> dict`

vmaas/reposcan/database/csaf_store.py
michalslomczynski: It should have been done in parser, not in CsafStore class. CsafStore class should just access the data and be done with it for good. When we will start using another product status ids, we will need to refactor CsafStore and the parser as well. Environment variable is not solving that technical debt. I can take care of that, but in separate PR and ticket.

vmaas/reposcan/database/csaf_store.py
michalslomczynski: it is used in refresh cache

vmaas/reposcan/database/csaf_store.py
michalslomczynski: Many people say the opposite is true.

vmaas/reposcan/database/csaf_store.py
psegedy: then do it in parser

vmaas/reposcan/database/csaf_store.py
psegedy: sorry I don't see this assigned anywhere, not even fetch param is used
>fetch – if True return the query results into a list (like in a [fetchall()](https://www.psycopg.org/docs/cursor.html#cursor.fetchall)). Useful for queries with RETURNING clause.
 I'm just referring to docs
I'm just curious, how does the result from returning from this query is used in refresh cache, is there some magic which assigns it to the dict?

vmaas/reposcan/database/csaf_store.py
michalslomczynski: I did it on purpose. I like the information that this function returns no rows to update.

vmaas/reposcan/database/csaf_store.py
michalslomczynski: Lets just stick to one style per repository at least. This code is just modified OVAL, something that existed prior to this PR.

vmaas/reposcan/database/csaf_store.py
psegedy: yes so return `return to_insert, None` 😄 so you don't even need to read the function and you know it returns None, but if you have strong preference then it's fine as it is, functionality is the same

vmaas/reposcan/database/csaf_store.py
psegedy: This code should not be just modified OVAL, this should be the code to store data about CSAF. OVAL contains data per repository and contains all updates and CVEs in a given repo. CSAF is CVE based, it contains information about given CVE and which products it affects. I accept that you were inspired by oval_store implementation, this is a separate module though, we don't need to repeat same mistakes.

vmaas/reposcan/database/csaf_store.py
michalslomczynski: Please try to be more professional in the future. Misunderstanding is a normal thing and can be resolved without being mean or sarcastic. Keep in mind your actions can negatively impact well being of others.

Those parameters are used in this line:
https://github.com/RedHatInsights/vmaas/blob/7f2134f71174ab68747127c0502f0bf1141d11e7/vmaas/reposcan/database/csaf_store.py#L109

psegedy: There was nothing mean, sarcastic or unprofessional in the comment, I'm sorry you understood it this way. I'm seriously asking how rows from returning are being populated, by 'magic' I mean python magic methods which are called on the background, sometimes as a side effect, it's a widely used term. I was asking if there is some magic that execute_values does that it populates dict, that's all.

Thank you for providing me the answer, now I see that you are getting these values using cur.fetchall(). There is a small pitfall with this approach though. cur.fetchall() will fetch only items from the last page. I think it would be safer to do it this way
```
rows = execute_values(cur, "insert into ... returning id, other_rows", fetch=True)
```

michalslomczynski: So you just answered yourself why `page_size=len(to_insert)`. It is done this way in the other parts of the code as well and it has been reviewed and approved by other team members. Maybe there is room for improvement, maybe there isn't. There are more important things to focus on right now, for example fetching product status id in parser, which was your job and was done incompletely. You would be able to fix that during the time you wasted on making those comments. Please prioritize your actions to align better with team objectives.

vmaas/reposcan/database/csaf_store.py
```diff
@@ -46,7 +49,76 @@ def _save_csaf_files(self, csaf_files: CsafFiles) -> list[int]:
...
+            self.conn.rollback()
+        finally:
+            cur.close()
+
     def store(self, csaf_data: CsafData):
         """Store collection of CSAF files into DB."""
         self._save_csaf_files(csaf_data.files)
-        # TODO: populate future objects in csaf store
+        self._populate_cves(csaf_data.cves)
+        self._delete_unreferenced_products()
```
psegedy, line 123: This has a nicer API with `store` method. However, it is called with each batch (it's called about 50 times with the first/full sync). Both are fine, mentioning in case if there are still some architecture design discussions are going on.

vmaas/reposcan/database/csaf_store.py
```diff
@@ -46,7 +49,76 @@ def _save_csaf_files(self, csaf_files: CsafFiles) -> list[int]:
...
+        product_tuples = [product.to_tuple() for product in product_list]
+        try:
+            cur = self.conn.cursor()
+            if product_tuples:
+                execute_values(cur, """
+                                    INSERT INTO csaf_products (cpe, package, module)
+                                    SELECT v.cpe, v.package, v.module
+                                    FROM (VALUES %s) AS v(cpe, package, module)
+                                    ON CONFLICT (cpe, package, module) DO NOTHING;
+                                    """, (product_tuples))
```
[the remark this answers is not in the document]
jdobes: It should be
```
INSERT INTO csaf_products (cpe, package, module)
VALUES %s
ON CONFLICT (cpe, package, module) DO NOTHING;
```

vmaas/reposcan/database/csaf_store.py
```diff
@@ -46,7 +49,76 @@ def _save_csaf_files(self, csaf_files: CsafFiles) -> list[int]:
...
+            cur = self.conn.cursor()
+            if product_tuples:
+                execute_values(cur, f"""
+                    WITH product_ids AS (
+                        SELECT id
+                        FROM csaf_products
+                        WHERE (cpe, package, module) IN %s
+                    )
+                    INSERT INTO csaf_cves (cve, csaf_product_id, product_status_id)
+                    SELECT '{cve}', id, {PRODUCT_STATUS_ID}
```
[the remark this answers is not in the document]
jdobes: Let's keep it as is, cve will be changed to foreign key later.

vmaas/reposcan/database/csaf_store.py
jdobes: I'm ok with both ways, this looks fine

vmaas/reposcan/database/csaf_store.py
```diff
@@ -46,7 +49,76 @@ def _save_csaf_files(self, csaf_files: CsafFiles) -> list[int]:
...
+            CSAF_FAILED_UPDATE.inc()
+            self.logger.exception("Failure while inserting or updating data to csaf_cves: %s", exc)
+            self.conn.rollback()
+        finally:
+            cur.close()
+
+    def _populate_cves(self, csaf_cves):
+        for cve, product_list in csaf_cves.items():
+            self._insert_missing_products(product_list)
+            self._insert_cves(cve, product_list)
```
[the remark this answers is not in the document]
jdobes: Yes, I agree, deleting obsolete rows from `csaf_cves` for a given CVE has to be done here (or inside `_insert_cves`). Otherwise `_delete_unreferenced_products` has no purpose.

Don't confuse it with deleting whole CVEs which can't be done yet.

jdobes (changes requested):

vmaas/reposcan/database/csaf_store.py
```diff
@@ -46,7 +48,125 @@ def _save_csaf_files(self, csaf_files: CsafFiles) -> list[int]:
...
+        finally:
+            cur.close()
+
+        return existing_products
+
+    def _get_product_rows(self, cve_products):
+        db_products = self._get_existing_products()
+
+        to_insert_products = cve_products - db_products
+        to_delete_products = db_products - cve_products
```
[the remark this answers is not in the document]
michalslomczynski: I forgot where cve clause in the select statement and it was supposed to be there. Thanks for the catch :+1:

vmaas/reposcan/database/csaf_store.py
```diff
@@ -46,7 +48,125 @@ def _save_csaf_files(self, csaf_files: CsafFiles) -> list[int]:
...
+    def _insert_missing_products(self, products_by_status: dict):
+        inserted_products = []
+        try:
+            cur = self.conn.cursor()
+            for status, products in products_by_status.items():
+                if products:
+                    ids = execute_values(cur, """
+                        INSERT INTO csaf_products (cpe, package, module)
+                        VALUES %s
+                        ON CONFLICT (cpe, package, module) DO NOTHING returning id;
```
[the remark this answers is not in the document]
michalslomczynski: I removed ON CONFLICT but then we get into trouble of unique constraint violation. Maybe empty string would make more sense there instead of NULL.

vmaas/reposcan/database/csaf_store.py
```diff
@@ -46,7 +48,125 @@ def _save_csaf_files(self, csaf_files: CsafFiles) -> list[int]:
...
         return db_ids
 
+    def _get_existing_cve_products(self, cve) -> Set[tuple]:
+        existing_products = set()
+        try:
+            cur = self.conn.cursor()
+            cur.execute("SELECT csaf_product_id FROM csaf_cves WHERE cve = %s", (cve,))
+            ids = [row[0] for row in cur.fetchall()]
+            if ids:
+                rows = execute_values(cur, "SELECT * from csaf_products where id in %s", (ids,), fetch=True)
```
jdobes, line 58: This should be `cur.execute` instead of `execute_values`.

vmaas/reposcan/database/csaf_store.py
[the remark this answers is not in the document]
jdobes: I see, if it complains about unique constraint violation, it indicates that there is some logic error earlier. I can look into it today what's happening here.

michalslomczynski:
Thanks, please do.

To me it looks like different CVEs might contain same CSAF products, and without querying whole csaf_products table we cannot know if it is already inserted for other CVEs.

michalslomczynski:
But maybe there are some better ways indeed. `ON CONFLICT DO NOTHING` while similar is probably faster.

vmaas/reposcan/database/csaf_store.py
```diff
@@ -46,7 +48,125 @@ def _save_csaf_files(self, csaf_files: CsafFiles) -> list[int]:
...
+            self.logger.exception("Failure while fetching CSAF products from the DB: %s", exc)
+        finally:
+            cur.close()
+
+        return existing_products
+
+    def _get_product_rows(self, cve, import_products):
+        cve_db_products = self._get_existing_cve_products(cve)
+
+        to_insert_products = import_products - cve_db_products
```
jdobes, line 73: This doesn't work correctly because format of these tuples is not same

`import_products` tuples are in format:
```
('cpe:/o:redhat:enterprise_linux:9', 'kernel-rt', 4, None)
```

but `cve_db_products` tuples are in format:
```
(141, 'cpe:/o:redhat:enterprise_linux:9', 'kernel-rt', None)
```

vmaas/reposcan/database/csaf_store.py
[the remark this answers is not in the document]
jdobes: Products to insert and to delete are not determined correctly, see above.

jdobes (changes requested):

vmaas/reposcan/database/csaf_store.py
```diff
@@ -46,7 +48,125 @@ def _save_csaf_files(self, csaf_files: CsafFiles) -> list[int]:
...
         return db_ids
 
+    def _get_existing_cve_products(self, cve) -> Set[tuple]:
+        existing_products = set()
+        try:
+            cur = self.conn.cursor()
+            cur.execute("SELECT csaf_product_id FROM csaf_cves WHERE cve = %s", (cve,))
+            ids = [row[0] for row in cur.fetchall()]
+            if ids:
+                cur.execute("SELECT * from csaf_products where id in %s", (tuple(ids),))
```
jdobes, line 58: Additionaly, let's merge these 2 selects into 1 using `JOIN`.

vmaas/reposcan/database/csaf_store.py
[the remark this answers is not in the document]
michalslomczynski: Should be fixed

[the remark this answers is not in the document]
jdobes: Now you should be able to remove the `ON CONFLICT`.

vmaas/reposcan/database/csaf_store.py
```diff
@@ -46,7 +48,125 @@ def _save_csaf_files(self, csaf_files: CsafFiles) -> list[int]:
...
+        return products_by_status
+
+    def _populate_cves(self, csaf_cves):
+        for cve in csaf_cves.keys():
+            cve_products = csaf_cves.to_set_tuples(cve, ["cpe", "package", "status_id", "module"])
+            to_insert_products, to_remove_product_ids = self._get_product_rows(cve_products)
+            cve_products_by_status = self._map_products_by_status(to_insert_products)
+            inserted_products = self._insert_missing_products(cve_products_by_status)
+            self._remove_cves(cve, to_remove_product_ids)
+            self._insert_cves(cve, inserted_products)
```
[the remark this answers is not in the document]
jdobes: This still needs to be fixed.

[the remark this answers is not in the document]
michalslomczynski: What I mean by that is for this particular case of `CVE-2023-1017`, `to_insert_products` and `to_delete_products` are empty.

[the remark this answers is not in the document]
psegedy: If I understand Jan's comment correctly, the following case does not seem to work (I haven't been testing the code so I might be totally wrong)
1. let's have CVE-1 with products a,b,c,d, and CVE-2 with products d,e,f
2. CVE-1 is being inserted so products a,b,c,d are inserted to DB and linked with CVE-1
3. CVE-2 is being inserted, products e,f are inserted and returned in `inserted_products`, these products (e,f) are linked with CVE-2, but product `d` is missed. Product `b` should be linked with CVE-2 as well but it isn't since it is not in `inserted_products`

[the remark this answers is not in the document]
jdobes: Yep, `inserted_products` contains only newly added rows. But some products may exist in DB already from previously synced CVEs and in this case they will not be included `inserted_products`. But they still need to be associated with currently processed CVE.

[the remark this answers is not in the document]
michalslomczynski: That is valid point indeed. I added code that handles it but we do not know if product is already in DB without querying whole csaf_products table since `cve_db_products = self._get_existing_cve_products(cve)`  uses
```
            cur.execute("""SELECT cp.cpe, cp.package, cp.module
                            FROM csaf_cves cc
                            JOIN csaf_products cp ON cc.csaf_product_id = cp.id
                            WHERE cc.cve = %s""", (cve,))
```
I guess I have to revert `WHERE cc.cve = %s` condition?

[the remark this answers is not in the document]
jdobes: I see, I think you need to do 2 selects:
- 1 for selecing all products in csaf_products to identify missing products. (Perhaps populate a dict at the beginning once for a whole batch (or sync) and keep selected values in memory + add newly inserted) and then use the dict to check what's missing.
- 1 for selecting only products for 1 CVE (like this) so you will not pass all possible products from DB to the `DELETE` statement.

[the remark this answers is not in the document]
michalslomczynski: Did something like this. Please take a look if its okay.

vmaas/reposcan/database/csaf_store.py
```diff
@@ -46,7 +47,132 @@ def _save_csaf_files(self, csaf_files: CsafFiles) -> list[int]:
...
 
+    def _get_product_rows(self, products_by_status):
+        to_associate_ids = []
+        to_insert = defaultdict(list)
+        try:
+            cur = self.conn.cursor()
+            for status, products in products_by_status.items():
+                if not products:
+                    continue
+                rows = execute_values(cur, """SELECT id, cpe, package, module
```
jdobes, line 58: The select works, except for a case when module is NULL.

```
SELECT id, cpe, package, module from csaf_products where (cpe, package, module) in (('cpe:/o:redhat:enterprise_linux:8', 'dotnet8.0', NULL));
```
returns nothing, even though such row is in the table

something like this could select correctly but I haven't tested it very well and it could be written nicer
```
rows = cur.execute("""SELECT id, cpe, package, module
                                      FROM csaf_products
                                      WHERE ((cpe, package, module) in %s AND module IS NOT NULL)
                                         OR ((cpe, package) in %s AND module IS NULL)""",
                                   (tuple([row for row in products]),
                                    tuple((cpe, package) for cpe, package, _ in products)))
```

vmaas/reposcan/database/csaf_store.py
```diff
@@ -46,7 +47,132 @@ def _save_csaf_files(self, csaf_files: CsafFiles) -> list[int]:
...
+        inserted_products = []
+        try:
+            cur = self.conn.cursor()
+            for status, products in products_by_status.items():
+                if products:
+                    ids = execute_values(cur, """
+                        INSERT INTO csaf_products (cpe, package, module)
+                        VALUES %s returning id;
+                        """, (products), fetch=True)
+                    inserted_products = [(result[0], status) for result in ids]
```
psegedy, line 90: only products from last iteration are added to `inserted_products`. I suppose it should be `inserted_products.extend([(result[0], status) for result in ids])`

vmaas/reposcan/database/csaf_store.py
```diff
@@ -46,7 +47,132 @@ def _save_csaf_files(self, csaf_files: CsafFiles) -> list[int]:
...
+            cur.close()
+
+    def _map_products_by_status(self, products):
+        products_by_status = defaultdict(list)
+
+        for product in products:
+            status_id = product[2]
+            product_data = (product[0], product[1], product[3])
+            if products_by_status[status_id] is None:
+                products_by_status[status_id] = []
```
psegedy, line 142: I think this is not necessary since defaultdict is used here, defaultdict returns the default value (list in this case) when key is not found. Is there a case when it can be None?

psegedy (commented):
Overall, the code looks good. I haven't been testing it, I'll leave it for Jan. Please take a look at inline comments.

vmaas/reposcan/database/csaf_store.py
michalslomczynski: Arguments have been simplified and I tested it against NULL modules and it worked.

PR #1076, opened 2024-02-19
fix(csaf): finalize db schema

psegedy:
RHINENG-7862

further changes will be added once #1073 is completed

### Complete Schema
csaf_schema_additions
(it's not `cve_detail` table but `cve`, I'll fix the diagram later, code should be correct)

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/database/upgrade_scripts/020-csaf_tables2.sql
+++ b/database/upgrade_scripts/020-csaf_tables2.sql
@@ -1,0 +1,68 @@
+DROP TABLE csaf_cves;
+DROP TABLE csaf_products;
+UPDATE csaf_file SET updated = NULL;
+
+-- -----------------------------------------------------
+-- Table vmaas.csaf_product
+-- -----------------------------------------------------
+CREATE TABLE IF NOT EXISTS csaf_product (
+  id                SERIAL,
+  cpe_id            INT NOT NULL,
+  package_name_id   INT NULL,
+  package_id        INT NULL,
+  module_stream     TEXT NULL CHECK (NOT empty(module_stream)),
+  PRIMARY KEY (id),
+  CONSTRAINT cpe_id
+    FOREIGN KEY (cpe_id)
+    REFERENCES cpe (id),
+  CONSTRAINT package_name_id
+    FOREIGN KEY (package_name_id)
+    REFERENCES package_name (id),
+  CONSTRAINT package_id
+    FOREIGN KEY (package_id)
+    REFERENCES package (id),
+  CONSTRAINT pkg_id CHECK(
+    (package_name_id IS NOT NULL AND package_id IS NULL) OR
+    (package_name_id IS NULL AND package_id IS NOT NULL))
+)TABLESPACE pg_default;
+
+CREATE UNIQUE INDEX ON csaf_product(cpe_id, package_name_id) WHERE package_name_id IS NOT NULL AND package_id IS NULL AND module_stream IS NULL;
+CREATE UNIQUE INDEX ON csaf_product(cpe_id, package_id) WHERE package_id IS NOT NULL AND package_name_id IS NULL AND module_stream IS NULL;
+CREATE UNIQUE INDEX ON csaf_product(cpe_id, package_name_id, module_stream) WHERE package_name_id IS NOT NULL AND package_id IS NULL AND module_stream IS NOT NULL;
+CREATE UNIQUE INDEX ON csaf_product(cpe_id, package_id, module_stream) WHERE package_id IS NOT NULL AND package_name_id IS NULL AND module_stream IS NOT NULL;
+
+-- -----------------------------------------------------
+-- Table vmaas.csaf_cve_product
+-- -----------------------------------------------------
+CREATE TABLE IF NOT EXISTS csaf_cve_product (
+  id                     SERIAL,
+  cve_id                 INT NOT NULL,
+  csaf_product_id        INT NOT NULL,
+  csaf_product_status_id INT NOT NULL,
+  csaf_file_id           INT NOT NULL,
+  PRIMARY KEY (id),
+  CONSTRAINT cve_id
+    FOREIGN KEY (cve_id)
+    REFERENCES cve (id),
+  CONSTRAINT csaf_product_id
+    FOREIGN KEY (csaf_product_id)
+    REFERENCES csaf_product (id),
+  CONSTRAINT csaf_product_status_id
+    FOREIGN KEY (csaf_product_status_id)
+    REFERENCES csaf_product_status (id),
+  CONSTRAINT csaf_file_id
+    FOREIGN KEY (csaf_file_id)
+    REFERENCES csaf_file (id),
+  CONSTRAINT csaf_cve_product_ids_uq
+    UNIQUE (cve_id, csaf_product_id)
+)TABLESPACE pg_default;
+
+
+-- -----------------------------------------------------
+-- vmaas users permission setup:
+-- vmaas_writer - has rights to INSERT/UPDATE/DELETE; used by reposcan
+-- vmaas_reader - has SELECT only; used by webapp
+-- -----------------------------------------------------
+GRANT SELECT, INSERT, UPDATE, DELETE ON ALL TABLES IN SCHEMA public TO vmaas_writer;
+GRANT USAGE, SELECT, UPDATE ON ALL SEQUENCES IN SCHEMA public TO vmaas_writer;
+GRANT SELECT ON ALL TABLES IN SCHEMA public TO vmaas_reader;
--- a/database/vmaas_db_postgresql.sql
+++ b/database/vmaas_db_postgresql.sql
@@ -16,7 +16,7 @@ CREATE TABLE IF NOT EXISTS db_version (
 )TABLESPACE pg_default;
 
 -- Increment this when editing this file
-INSERT INTO db_version (name, version) VALUES ('schema_version', 19);
+INSERT INTO db_version (name, version) VALUES ('schema_version', 20);
 
 -- -----------------------------------------------------
 -- evr type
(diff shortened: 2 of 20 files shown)


vmaas/reposcan/database/csaf_store.py
```diff
@@ -28,151 +56,327 @@ def delete_csaf_file(self, name: str):
...
-                                WHERE c.csaf_product_id IS NULL)""")
+                                FROM csaf_product AS p
+                                LEFT JOIN csaf_cve_product AS c ON c.csaf_product_id = p.id
+                                WHERE c.csaf_product_id IS NULL)"""
+            )
             self.conn.commit()
         except Exception as exc:
             CSAF_FAILED_DELETE.inc()
-            self.logger.exception("Failure while deleting data from csaf_products: %s", exc)
+            self.logger.exception("Failure while deleting data from csaf_product: %s", exc)
```
jdobes, line 373: incorrect usage of `logger.exception()` (was before already)

exception is automatically added to the log message

```
self.logger.exception("Failure while deleting data from csaf_products: ")
```

vmaas/reposcan/database/csaf_store.py
```diff
@@ -28,151 +56,327 @@ def delete_csaf_file(self, name: str):
...
+                cur,
+                """
+                    INSERT INTO csaf_cve_product (cve_id, csaf_product_id, csaf_product_status_id, csaf_file_id)
                     VALUES %s
-                    ON CONFLICT (cve, csaf_product_id)
-                    DO UPDATE SET product_status_id = EXCLUDED.product_status_id;
-                    """, [(cve, product[0], product[1]) for product in products])
-                self.conn.commit()
+                    ON CONFLICT (cve_id, csaf_product_id)
+                    DO UPDATE SET csaf_product_id = EXCLUDED.csaf_product_id
```
jdobes, line 306: It was previously `DO UPDATE SET product_status_id = EXCLUDED.product_status_id`, is this an error?

vmaas/reposcan/database/csaf_store.py
```diff
@@ -28,151 +56,327 @@ def delete_csaf_file(self, name: str):
...
+        except Exception as exc:
+            CSAF_FAILED_IMPORT.inc()
+            self.logger.exception("Failed to import csaf file to DB: %s", exc)
+            raise CsafStoreException("Failed to import csaf file to DB")
+        finally:
+            cur.close()
+        for row in rows:
+            # translate file name to cve, e.g. 2023/cve-2023-0030.json -> CVE-2023-0030
+            match = re.search(CVE_PATTERN, row[1])
+            if match:
```
jdobes, line 86: What happens when there is a valid CVE file in repo which name doesn't match? I'd not rely on file name, rather extract the CVE name from inside the file.

jdobes (changes requested):

database/vmaas_db_postgresql.sql
```diff
@@ -1307,34 +1307,59 @@ ON CONFLICT DO NOTHING;
...
+    FOREIGN KEY (package_name_id)
+    REFERENCES package_name (id),
+  CONSTRAINT package_id
+    FOREIGN KEY (package_id)
+    REFERENCES package (id),
+  CONSTRAINT module_stream_id
+    FOREIGN KEY (module_stream_id)
+    REFERENCES module_stream (id),
+  CONSTRAINT csaf_product_ids_uq
+    UNIQUE (cpe_id, package_name_id, package_id, module_stream_id),
```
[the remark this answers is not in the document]
psegedy: good catch, thanks

vmaas/reposcan/database/csaf_store.py
psegedy: okay, I'll fix them

vmaas/reposcan/database/csaf_store.py
psegedy: yes, that's an error, thanks

database/vmaas_db_postgresql.sql
```diff
@@ -1307,34 +1307,61 @@ ON CONFLICT DO NOTHING;
...
+    REFERENCES cpe (id),
+  CONSTRAINT package_name_id
+    FOREIGN KEY (package_name_id)
+    REFERENCES package_name (id),
+  CONSTRAINT package_id
+    FOREIGN KEY (package_id)
+    REFERENCES package (id),
+  CONSTRAINT module_stream_id
+    FOREIGN KEY (module_stream_id)
+    REFERENCES module_stream (id),
```
jdobes, line 1330: I'm starting to think against using FK of `module_stream` table. Problem is that from CSAF we get only the module name and module_stream_name, while the `module_stream` table contains different and more granular data. We're unable to match any row from `module_stream` precisely.

`module_stream` table data implies also:
- architecture
- stream version
- stream context

To have an idea how many different rows can be matched for a single `module`:`module_stream` pair:

```
select m.name, ms.stream_name, count(*) from module m join module_stream ms on m.id = ms.module_id group by m.name, ms.stream_name limit 10;
```

```
 NAME               | STREAM_NAME | COUNT
--------------------+-------------+-------
 php                | 8.2         | 12
 nginx              | 1.16        | 288
 rhn-tools          | 1.0         | 477
 scala              | 2.1         | 126
 log4j              | 2           | 90
 satellite-5-client | 1.0         | 357
 ruby               | 2.6         | 324
 postgresql         | 13          | 506
 perl-YAML          | 1.24        | 1284
 mariadb-devel      | 10.3        | 160
```

jdobes (changes requested):

database/vmaas_db_postgresql.sql
jdobes: We could either store the module information directly as `TEXT` or create some simplified module table with only unique `module:ms` string pairs.

vmaas/reposcan/database/csaf_store.py
```diff
@@ -1,178 +1,378 @@
...
         finally:
             cur.close()
 
-    def _map_products_by_status(self, products) -> dict[int, list[tuple[str, str, str]]]:
-        products_by_status = defaultdict(list)
+    def _populate_cves(self, csaf_cves: model.CsafCves) -> None:
+        for cve, products in csaf_cves.items():
+            products_copy = deepcopy(products)  # only for logging of failed cves
+            try:
+                self._update_product_ids(products)
```
jdobes, line 345: Trying with empty database and after this function is called, `products` is always empty list which results that no products are ever inserted on the next line.

vmaas/reposcan/database/csaf_store.py
```diff
@@ -1,178 +1,378 @@
...
+            self.logger.exception("%s: ", FAILED_PRODUCT_DELETE)
             self.conn.rollback()
+            raise CsafStoreException(FAILED_PRODUCT_DELETE) from exc
         finally:
             cur.close()
 
-    def store(self, csaf_data: CsafData):
+    def store(self, csaf_data: model.CsafData) -> None:
         """Store collection of CSAF files into DB."""
         self._save_csaf_files(csaf_data.files)
```
jdobes, line 377: Not really related to the PR but could the `updated` timestamp be updated for each file separately and only once the CVE data are succefully synced (no error occurs)?

jdobes (changes requested):

database/upgrade_scripts/020-csaf_tables2.sql
```diff
@@ -0,0 +1,67 @@
+DROP TABLE csaf_cves;
+DROP TABLE csaf_products;
+
```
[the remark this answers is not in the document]
psegedy: we have env var to trigger re-sync, is this necessary?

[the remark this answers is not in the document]
jdobes: well, this will save 4 MRs (to enable and disable the flag x 2 environments)

vmaas/reposcan/database/csaf_store.py
psegedy: did you have any cpes and packages in db before trying the sync? if cpe for the product is missing in db then the product is skipped, cpes are not inserted as part of the csaf sync. Same for the packages, if the package is missing then the product (and CVE) is skipped

database/upgrade_scripts/020-csaf_tables2.sql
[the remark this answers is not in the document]
jdobes: my idea is that the env var should be used when there is no schema change so we don't add a migration doing only this

but since there is a schema change that implies need to re-sync, we can do it directly in the migration

vmaas/reposcan/database/csaf_store.py
psegedy: sure why not

database/upgrade_scripts/020-csaf_tables2.sql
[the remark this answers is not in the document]
psegedy: okay

vmaas/reposcan/database/csaf_store.py
jdobes: I've synced CPEs but seems they are not matched... Looking

jdobes: Discussing privately, issue seems more complicated

psegedy:
We will need to update cvemap for ephemeral because cves from csaf are missing in the current cvemap used for testing

psegedy:
Based on logs we still have an issue - it tried to insert duplicate row, I will investigate
```
2024-03-20 11:39:50,952:ERROR:vmaas.reposcan.database.object_store:Failure while inserting data into csaf_product: 'Traceback (most recent call last):\n  File "/vmaas/vmaas/reposcan/database/csaf_store.py", line 262, in _insert_missing_products\n    rows = execute_values(\n           ^^^^^^^^^^^^^^^\n  File "/usr/local/lib64/python3.11/site-packages/psycopg2/extras.py", line 1299, in execute_values\n    cur.execute(b\'\'.join(parts))\npsycopg2.errors.UniqueViolation: duplicate key value violates unique constraint "csaf_product_cpe_id_package_name_id_idx"\nDETAIL:  Key (cpe_id, package_name_id)=(907, 12550) already exists.\n'|
```

jdobes (approved):

PR #1079, opened 2024-02-21
fix(csaf): parse product status

psegedy:

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/vmaas/reposcan/redhatcsaf/modeling.py
+++ b/vmaas/reposcan/redhatcsaf/modeling.py
@@ -7,6 +7,7 @@ import csv
 from dataclasses import dataclass
 from dataclasses import field
 from datetime import datetime
+from enum import IntEnum
 from typing import Iterator
 from typing import Optional
 from typing import overload
@@ -14,6 +15,18 @@ from typing import overload
 import attr
 
 from vmaas.common.date_utils import parse_datetime
+
+
+class CsafProductStatus(IntEnum):
+    """CSAF product status enum."""
+    FIRST_AFFECTED = 1
+    FIRST_FIXED = 2
+    FIXED = 3
+    KNOWN_AFFECTED = 4
+    KNOWN_NOT_AFFECTED = 5
+    LAST_AFFECTED = 6
+    RECOMMENDED = 7
+    UNDER_INVESTIGATION = 8
 
 
 @dataclass
@@ -125,6 +138,7 @@ class CsafProduct:
 
     cpe: str
     package: str
+    status_id: int
     module: Optional[str] = None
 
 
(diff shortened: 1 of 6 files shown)


vmaas/reposcan/redhatcsaf/modeling.py
```diff
@@ -125,6 +138,7 @@ class CsafProduct:
 
     cpe: str
     package: str
+    status_id: int
```
michalslomczynski, line 141: You could also consider original CSAF structure approach and include status information in CsafCves class `_cves: dict[str, dict[CsafProductStatus, list[CsafProduct]]] = attr.Factory(dict)`.

To have optimal performance Csaf Store will probably need to insert missing products in separate inserts per CsafProductStatus. Currently it needs to iterate over every CsafProduct to make such dict. It is because we do not have `status_id` column in `csaf_products` table and association of `id (csaf_products)` and `status_id` is needed to insert CVEs.

michalslomczynski: There is solution to this like
```
    def _insert_missing_products(self, products): # (cpe, package, status, module)
                ...
                execute_values(cur, """
                    WITH input_data (cpe, package, module, status_id) AS (
                        VALUES %s
                    )
                    INSERT INTO csaf_products (cpe, package, module)
                    SELECT cpe, package, module
                    FROM input_data
                    ON CONFLICT (cpe, package, module) DO NOTHING
                    RETURNING id,
                    (SELECT ARRAY(SELECT status_id FROM input_data WHERE csaf_products.cpe = input_data.cpe));
                    """, [(product[0], product[1], product[3], product[2]) for product in products])
                inserted_products = [(result[0], result[1][0]) for result in cur.fetchall()]
```
but I think it is suboptimal.

psegedy: Yes it is also a possibility to have `dict[str, dict[CsafProductStatus, list[CsafProduct]]]` but it might complicate parsing with no added benefit. `status_id` can't be included in `csaf_products` table to avoid M:N relations and to have DB schema in normal form. Current schema is not good, we will change it. That dict (and iteration over every CsafProduct) is not needed when inserting to `csaf_products`, the illustrated insert can be as easy as

```
inserted = execute_values(cur, """
   INSERT INTO csaf_products (cpe, package, module) VALUES %s RETURNING id
   """, products, fetch=True
  )
```
since the function should insert only missing products (no on conflict needed) and status_id does not need to be included in tuples in `products` (it is possible to select which attributes to include when calling to_tuples)  argument since it is not inserted to this table.

PR #1081, opened 2024-03-12
feat(reposcan): Add CSAF SQLite dump

michalslomczynski:
RHINENG-7868

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management


michalslomczynski:
DB file dumps. vmaas-lib reads tables and rows.

michalslomczynski:
Still a bit of WIP due to vmaas-lib changes, but maybe it could be merged at this point.

vmaas/reposcan/exporter.py
```diff
@@ -748,6 +749,59 @@ def _dump_oval(self, dump):
...
+                dump.execute("INSERT INTO csaf_product VALUES (?, ?, ?, ?, ?)",
+                             (csaf_product_id, cpe_id, package_name_id, package_id, module_stream))
+
+        # Dump data from csaf_cve_product table
+        dump.execute("""CREATE TABLE IF NOT EXISTS csaf_cve_product (
+                        id INTEGER PRIMARY KEY AUTOINCREMENT,
+                        cve_id INTEGER NOT NULL,
+                        csaf_product_id INTEGER NOT NULL,
+                        csaf_product_status_id INTEGER NOT NULL,
+                        csaf_file_id INTEGER NOT NULL)
```
psegedy, line 775: Is `csaf_file_id` needed for evaluation? We can exclude it from the dump if it won't be used

vmaas/reposcan/exporter.py
```diff
@@ -748,6 +749,59 @@ def _dump_oval(self, dump):
...
+                            FROM csaf_cve_product""")
+            for csaf_cve_id, cve_id, csaf_product_id, csaf_product_status_id, csaf_file_id in cursor:
+                dump.execute("INSERT INTO csaf_cve_product VALUES (?, ?, ?, ?, ?)",
+                             (csaf_cve_id, cve_id, csaf_product_id, csaf_product_status_id, csaf_file_id))
+
+        # Dump data from csaf_file table
+        dump.execute("""CREATE TABLE IF NOT EXISTS csaf_file (
+                        id INTEGER PRIMARY KEY AUTOINCREMENT,
+                        name TEXT UNIQUE NOT NULL,
+                        updated TIMESTAMP)""")
```
psegedy, line 788: and if csaf_file_id is not needed, then we can exclude the whole csaf_file table from the dump

vmaas/reposcan/exporter.py
```diff
@@ -748,6 +749,59 @@ def _dump_oval(self, dump):
...
+            for csaf_product_id, cpe_id, package_name_id, package_id, module_stream in cursor:
+                dump.execute("INSERT INTO csaf_product VALUES (?, ?, ?, ?, ?)",
+                             (csaf_product_id, cpe_id, package_name_id, package_id, module_stream))
+
+        # Dump data from csaf_cve_product table
+        dump.execute("""CREATE TABLE IF NOT EXISTS csaf_cve_product (
+                        id INTEGER PRIMARY KEY AUTOINCREMENT,
+                        cve_id INTEGER NOT NULL,
+                        csaf_product_id INTEGER NOT NULL,
+                        csaf_product_status_id INTEGER NOT NULL,
```
psegedy, line 774: It would be beneficial to have `cve` string instead of `cve_id` in the map in vmaas-lib so we don't need to translate id to name during the evaluation. I had a quick look and it seems that the current vmaas-lib code does use only cve strings. It can be done either here by replacing `cve_id INTEGER NOT NULL` with `cve TEXT NOT NULL` and joining with `cve` table before the insert, or it can be done during the load in vmaas-lib.

vmaas/reposcan/exporter.py
```diff
@@ -748,6 +749,59 @@ def _dump_oval(self, dump):
...
+        # Dump data from csaf_product table
+        dump.execute("""CREATE TABLE IF NOT EXISTS csaf_product (
+                        id INTEGER PRIMARY KEY AUTOINCREMENT,
+                        cpe_id INTEGER NOT NULL,
+                        package_name_id INTEGER,
+                        package_id INTEGER,
+                        module_stream_id INTEGER
+                    )""")
+        with self._named_cursor() as cursor:
+            cursor.execute("""SELECT *
```
psegedy, line 763: Even though we want to include all columns in the dump now, it might be better to list which columns we want to select. We might rename some columns or add more columns to tables in the future. Listing desired columns will guarantee their order. I'd list columns in all selects.

psegedy (changes requested):

vmaas/reposcan/exporter.py
michalslomczynski: I haven't read whole vmaas-lib repo, just took a look at first existing functions and it operates with cve_id then translates it to cve name in cache `CveNames  map[int]string`. Then cve details are accessed in another cache `CveDetail map[string]CveDetail`.

vmaas/reposcan/exporter.py
michalslomczynski: No, it is old code. I'll remove.

vmaas/reposcan/exporter.py
psegedy: It will work, there's no problem. I just though we might want to use the CVE string directly so we won't need to add another for loop for every list of cve_ids. I was just looking at how current evaluation works and it seems to be using cve names directly instead of returning ids and then converting it to cve names. We don't seem to be operating with cve_ids at all in vmaas-lib, only with cve names.

vmaas/reposcan/exporter.py
```diff
@@ -748,6 +749,47 @@ def _dump_oval(self, dump):
...
 
+    def _dump_csaf(self, dump):
+        """Select CSAF data"""
+        # Dump data from csaf_product table
+        dump.execute("""CREATE TABLE IF NOT EXISTS csaf_product (
+                        id INTEGER PRIMARY KEY AUTOINCREMENT,
+                        cpe_id INTEGER NOT NULL,
+                        package_name_id INTEGER DEFAULT 0,
+                        package_id INTEGER DEFAULT 0,
+                        module_stream TEXT DEFAULT ''
```
psegedy, line 760: these defaults and coalesce are because sql.Scan in go can't load NULL to int/string? I think we did it for OVAL the other way around, added coalesce to select in go code. Both solutions are fine with me.

michalslomczynski: Yes. It returned ~`can't parse NULL to int` error without it. For me it is no difference either.

psegedy (approved):

PR #1101, opened 2024-05-23
Enable Platform-Security GitHub Workflow

casey-williams-rh:
## Overview
Enabling the `Platform Security GitHub Workflow` - This workflow provides Red Hat ConsoleDot Teams with a way to scan the containers they create in a convenient, automated, and reliable manner within their GitHub repository. The Platform Security Github Workflow lets teams get security feedback as they open Pull Requests.

REF: https://github.com/RedHatInsights/platform-security-gh-workflow

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/.github/workflows/security-scanning.yml
+++ b/.github/workflows/security-scanning.yml
@@ -1,0 +1,49 @@
+# This workflow uses actions that are not certified by GitHub.
+# They are provided by a third-party and are governed by
+# separate terms of service, privacy policy, and support
+# documentation.
+
+# The default values used in the docker build commands are the root
+# directory '.' and the dockerfile name of 'Dockerfile'. If there is 
+# a need to change these do so in your local workflow template (this file) and
+# change them there. HINT: Look at the bottom of this file.
+
+# This workflow checks out code, builds an image, performs a container image
+# vulnerability scan with Anchore's Grype tool, and generates an
+# SBOM via Anchore's Syft tool
+
+# For more information on Anchore's container image scanning tool Grype, see
+# https://github.com/anchore/grype
+
+# For more information about the Anchore SBOM tool, Syft, see
+# https://github.com/anchore/syft
+
+name: ConsoleDot Platform Security Scan
+
+on:
+  push:
+    branches: [ "main", "master" ]
+  pull_request:
+    # The branches below must be a subset of the branches above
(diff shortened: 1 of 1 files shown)


psegedy (changes requested):
@casey-williams-rh this security scan is not ideal. First of all, it uses ubi-8 repos while the image built in for stage/prod is build with rhel-8 repos, which means that different packages could be installed. Shouldn't this job run on ci.int.devshift.net on a rhsm registered host?
Anchore-Grype-Vulnerability-San job reports that `setuptool` is in old version which is not true
```
setuptools              39.2.0                           65.5.1                                   python  GHSA-r9hx-vwmv-q579  High
```
this is the version used by platform-python which cannot be updated and it also cannot be removed because `dnf` depends on it. In fact our application uses python3.11 with setuptools>=69

Are there any plans to improve those jobs? It will always fail and when something always fails it will be ignored by people which I think is not what we want.

casey-williams-rh:
@psegedy, Upon further review of the Dockerfile, I understand what you mean regarding _"This workflow uses ubi-8 repos while the image built for stage/prod is build with rhel-8 repos"._

Because the job is not running on a registered RHEL system, alternative packages are being pulled in. 

```
RUN FULL_RHEL=$(microdnf repolist --enabled | grep rhel-8) ; \
    if [ -z "$FULL_RHEL" ] ; then \
        rpm -Uvh http://mirror.centos.org/centos/8-stream/BaseOS/x86_64/os/Packages/centos-stream-repos-8-4.el8.noarch.rpm \
                 http://mirror.centos.org/centos/8-stream/BaseOS/x86_64/os/Packages/centos-gpg-keys-8-4.el8.noarch.rpm && \
        sed -i 's/^\(enabled.*\)/\1\npriority=200/;' /etc/yum.repos.d/CentOS*.repo ; \
    fi
```

casey-williams-rh:
@psegedy , I'm closing this PR for now so my team and I can better evaluate how we can scan the VMAAS Container Image.

PR #1193, opened 2024-11-15
feat(webapp): add go controller for cves endpoint

Dugowitch:

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

### DIFF
--- a/deploy/clowdapp.yaml
+++ b/deploy/clowdapp.yaml
@@ -51,6 +51,8 @@ objects:
           value: ${NEWER_RELEASEVER_REPOS}
         - name: NEWER_RELEASEVER_CSAF
           value: ${NEWER_RELEASEVER_CSAF}
+        - name: ENABLE_GO_CVES
+          value: ${ENABLE_GO_CVES}
         resources:
           limits:
             cpu: ${CPU_LIMIT_WEBAPP_GO}
@@ -494,3 +496,6 @@ parameters:
   value: "true"
 - name: DB_DUMP_BUCKET
   value: insights-vmaas-dump-storage
+- name: ENABLE_GO_CVES
+  description: Enable go implementation of the cves endpoint
+  value: "false"
--- a/vmaas-go/webapp/controllers/cves.go
+++ b/vmaas-go/webapp/controllers/cves.go
@@ -1,0 +1,44 @@
+package controllers
+
+import (
+	"net/http"
+
+	"github.com/gin-gonic/gin"
+	"github.com/redhatinsights/vmaas-lib/vmaas"
+	"github.com/redhatinsights/vmaas/base/core"
+	"github.com/redhatinsights/vmaas/base/utils"
+)
+
+func CvesHandler(c *gin.Context) {
+	if !isCacheLoaded(c) {
+		return
+	}
+	cve := c.Param("cve")
+	req := vmaas.CvesRequest{Cves: []string{cve}}
+
+	res, err := core.VmaasAPI.Cves(&req)
+	if err != nil {
+		utils.LogAndRespError(c, err)
+		return
+	}
+	c.JSON(http.StatusOK, res)
+}
+
+func CvesPostHandler(c *gin.Context) {
+	if !isCacheLoaded(c) {
+		return
+	}
+	req := vmaas.CvesRequest{}
+	err := bindValidateJSON(c, &req)
+	if err != nil {
+		utils.LogAndRespBadRequest(c, err)
+		return
+	}
+
+	cves, err := core.VmaasAPI.Cves(&req)
+	if err != nil {
+		utils.LogAndRespError(c, err)
+		return
+	}
+	c.JSON(http.StatusOK, cves)
+}
--- a/vmaas-go/webapp/routes/routes.go
+++ b/vmaas-go/webapp/routes/routes.go
@@ -3,6 +3,7 @@ package routes
 import (
 	"github.com/gin-contrib/gzip"
 	"github.com/gin-gonic/gin"
+	"github.com/redhatinsights/vmaas/base/utils"
 	"github.com/redhatinsights/vmaas/webapp/controllers"
 )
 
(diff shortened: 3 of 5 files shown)


deploy/clowdapp.yaml
```diff
@@ -51,6 +51,8 @@ objects:
           value: ${NEWER_RELEASEVER_REPOS}
         - name: NEWER_RELEASEVER_CSAF
           value: ${NEWER_RELEASEVER_CSAF}
+        - name: ENABLE_GO_IMPLEMENTATION
+          value: ${ENABLE_GO_IMPLEMENTATION}
```
psegedy, line 55: make separate env for every endpoint so we can enable them in prod gradually once they are tested, change the name to something like ENABLE_GO_CVES

vmaas-go/webapp/controllers/cves.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/redhatinsights/vmaas-lib/vmaas"
+	"github.com/redhatinsights/vmaas/base/core"
+	"github.com/redhatinsights/vmaas/base/utils"
+)
+
+func CvesHandler(c *gin.Context) {
+	if !isCacheLoaded(c) {
+		return
+	}
+	cves := strings.Split(c.Param("cves"), ",")
```
psegedy, line 17: GET handler shows results for only 1 CVE, don't split the string, pass what is in cves

psegedy (changes requested):

vmaas-go/webapp/routes/routes.go
```diff
@@ -12,4 +13,9 @@ func InitAPI(api *gin.RouterGroup) {
 	api.POST("/updates", controllers.UpdatesPostHandler)
 	api.GET("/vulnerabilities/:package", controllers.VulnerabilitiesHandler)
 	api.POST("/vulnerabilities", controllers.VulnerabilitiesPostHandler)
+
+	if utils.Cfg.EnableGoCves {
+		api.GET("/cves/:cve")
+		api.POST("/cves")
```
psegedy, line 19: controller is missing

psegedy (changes requested):

vmaas-go/base/utils/gin.go
```diff
@@ -105,6 +106,25 @@ func respStatusError(c *gin.Context, code int, err error) {
...
+	if strings.HasSuffix(errMessage, "looking for beginning of value") {
+		parts := strings.Split(errMessage, ` `)
+		if len(parts) < 3 {
+			return errors.New("malformed input")
+		}
+		return errors.New("malformed input: invalid character " + parts[2])
+	}
+	return err
+}
+
```
psegedy, line 127: Parsing errors based on error message and then changing the error message is not optimal. I'd prefer wrapping original error.
Error for wrong date can be for example handled when UnmarshalJSON for time fails (it would require custom type for time so UnmarshalJSON can be defined, something like https://github.com/RedHatInsights/patchman-engine/blob/master/base/types/timestamp.go). Anyways, timestamp is parsed only by `bindValidateJSON` so at least move the current logic to that function.

`looking for beginning of value` is returned by regex error? In that case you can wrap the error in vmaas-lib before returning regex error or return some custom error type and then change the message after checking the error type with errors.Is

vmaas-go/base/utils/gin.go
```diff
@@ -105,6 +106,25 @@ func respStatusError(c *gin.Context, code int, err error) {
...
 }
 
+func processBadRequestErrMessage(err error) error {
+	errMessage := err.Error()
+	if strings.HasPrefix(errMessage, "parsing time") {
+		parts := strings.Split(errMessage, `"`)
+		if len(parts) < 2 {
+			return errors.New("Wrong date format (not ISO format with timezone)")
+		}
+		return errors.New("Wrong date format (not ISO format with timezone): " + parts[1])
```
psegedy, line 116: what's the purpose of splitting the error message? When it has only 1 `"` then the error will be `Wrong date format (not ISO format with timezone` and when there are more `"` then the second part is wrapped by the new message and other parts are discarded?
If the point is to only wrap the current error, you can use `errors.Wrap` from `github.com/pkg/errors` which is already used in this module or https://pkg.go.dev/errors#Join

psegedy (changes requested):

vmaas-go/base/utils/gin.go
Dugowitch: The purpose is to get the original timestamp being parsed, the error message without one is just a fallback.

vmaas-go/base/utils/gin.go
Dugowitch: The original error `parsing time "2017-01-01T00:00:00" as "2006-01-02T15:04:05Z07:00": cannot parse "" as "Z07:00"` is clunky and requires some go-specific knowledge to understand. That's why I opted to replace it rather than to just wrap it.

`invalid character 'N' looking for beginning of value` also occurs during JSON parsing, for example, when there is `NaN` instead of a number.

I will move the logic. Should I also wrap the errors instead?

psegedy: ok, let's leave it as it is even though `invalid character 'N' looking for beginning of value` won't say much to the user of the API, we can fix that later in vmaas-lib if it becomes a problem. We need to make sure tests are passing so we can merge this

psegedy (changes requested):
`tests.test_json.TestJSON.test_json_cves[empty_json]` fails because app returns EOF instead of expected error message `Request body must not be empty`

 `tests.test_cves.TestCVEsCorrect.test_get[]`, request is probably something like `GET /api/v3/cves/''` and GIN might treat it as missing path parameter so it returns 405, please discuss with Max what to do with it, whether app should still return 200 or if tests should expect 405 in this case

psegedy (approved):

PR #1810, opened 2026-05-21
refactor: use plain golang http server instead of nginx

jdobes:
RHINENG-26574

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

## Summary by Sourcery

Replace nginx-based reposcan data serving with an embedded Go HTTP file server and remove nginx from the container image and RPM inputs.

New Features:
- Introduce a Go-based reposcan data HTTP server with request logging middleware.

Enhancements:
- Wire the new reposcan data server into the main Go entrypoint and container entrypoint script to run alongside the reposcan service.
- Update kernel and python-pip RPM locks to newer RHEL 9.8 builds and refresh associated module metadata.

Build:
- Remove nginx and related assets from RPM manifests and Docker image dependencies to simplify the runtime stack.


entrypoint.sh
```diff
@@ -26,10 +26,8 @@ if [[ ! -z $1 ]]; then
         cd go/src/vmaas
         exec ./main webapp
     elif [[ "$1" == "reposcan" ]]; then
+        /vmaas/go/src/vmaas/main reposcan-data-server &
```
vkrizan, line 25: Could there be any error handling if this does not start correctly?

vmaas-go/reposcan_data_server/reposcan_data_server.go
```diff
@@ -0,0 +1,67 @@
...
+	n, err := rw.ResponseWriter.Write(b)
+	rw.bytes += n
+	return n, err
+}
+
+// loggingMiddleware logs each request with method, path, status, bytes, and duration.
+func loggingMiddleware(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		tStart := time.Now()
+		rw := &responseWriter{ResponseWriter: w, status: http.StatusOK}
```
vkrizan, line 35: The `status` of the `responseWriter` will always be set to StatusOK, as the `next.ServeHTTP()` have no means to set it: it is not part of the interface, is is not handled by the custom RW and is private attribute.

jdobes: I've checked the functionality and the status code is re-written by the http server using `WriteHeader` method for different than HTTP 200 responses

entrypoint.sh
jdobes: I've extended reposcan healthcheck to check status of data server as well

vmaas-go/reposcan_data_server/reposcan_data_server.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+func (rw *responseWriter) Write(b []byte) (int, error) {
+	n, err := rw.ResponseWriter.Write(b)
+	rw.bytes += n
+	return n, err
+}
+
+// loggingMiddleware logs each request with method, path, status, bytes, and duration.
+func loggingMiddleware(next http.Handler) http.Handler {
```
vkrizan, line 32: There needs to be an a different way for adding an access log that would not require a custom middleware.

jdobes: There is no built-in access logging in `net/http`, the only other option is to use third-party logging middleware implementation or whole http framework like gin

vkrizan (dismissed):

vkrizan (approved):

PR #1825, opened 2026-05-29
chore(deps): bump starlette to 1.0.1

Odilhao:
Bumps starlette to 1.0.1 to fix CVE-2026-48710.

**CVE:** CVE-2026-48710 — Starlette: Security restriction bypass via malformed HTTP Host header. Missing Host header validation poisons `request.url.path`, allowing path-based security checks to be bypassed.

**Fix version:** starlette 1.0.1
**Advisory:** https://github.com/Kludex/starlette/security/advisories/GHSA-86qp-5c8j-p5mr


jdobes (commented):
Seems it's not covered by tests here but I'm worried that there might be the same regression with Swagger as in https://github.com/RedHatInsights/vulnerability-engine/pull/2367

Odilhao:
Good catch — the Swagger UI regression on vulnerability-engine foreman branches (PR #2367, #2368) was caused by connexion 3.1.0's runtime incompatibility with starlette 1.0.x. vmaas is not affected by the same issue: vmaas uses starlette as a direct dependency (not via connexion's Swagger UI path), and the failing test in vulnerability-engine specifically exercises connexion's `/openapi.json` endpoint after a `$ref` request body call (issue #2029, now closed). The vulnerability-engine PRs have been updated to also bump connexion to 3.3.0.

jdobes (approved):

PR #1941, opened 2026-09-04
Customize iop-vmaas-sat-6-20 build pipelines

brantleyr:
Switch the iop-vmaas-sat-6-20 pull-request and push PipelineRuns from the generated inline pipeline to the shared `docker-build-oci-ta` pipeline (v1.73.0), matching the other iop-vmaas release branches.

- Enable hermetic builds with pip/rpm/gomod prefetch (matches the repo's existing hermetic components).
- Build source image and set release labels/CPE for satellite/iop-vmaas-rhel9 (6.20, el9).
- Bump prefetch/SAST task memory to 4Gi, as used by the current build pipelines.

This resolves the enterprise-contract failure on foreman-5.0, which came from the generated pipeline building non-hermetically.

## Summary by Sourcery

Update the iop-vmaas Satellite 6.20 build pipelines to use the shared hermetic OCI pipeline and release-specific image configuration.

Bug Fixes:
- Prevent enterprise-contract failures by replacing the non-hermetic generated build pipeline with a trusted shared pipeline.

Enhancements:
- Configure both pull-request and push builds to use the shared docker-build-oci-ta pipeline with hermetic dependency prefetching for pip, RPM, and Go modules.
- Produce source images with Satellite IoP Vmaas RHEL 9 release metadata and CPE labels.
- Increase prefetch and SAST task memory allocations to 4Gi.

### DIFF
--- a/.tekton/iop-vmaas-sat-6-20-pull-request.yaml
+++ b/.tekton/iop-vmaas-sat-6-20-pull-request.yaml
@@ -10,6 +10,8 @@ metadata:
     pipelinesascode.tekton.dev/max-keep-runs: "3"
     pipelinesascode.tekton.dev/on-cel-expression: event == "pull_request" && target_branch
       == "foreman-5.0"
+    pipelinesascode.tekton.dev/pipeline: https://github.com/RedHatInsights/konflux-pipelines/raw/v1.73.0/pipelines/docker-build-oci-ta.yaml
+  creationTimestamp: null
   labels:
     appstudio.openshift.io/application: iop-vmaas-sat-6-20
     appstudio.openshift.io/component: iop-vmaas-sat-6-20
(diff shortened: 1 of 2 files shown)


brantleyr:
The failing `test_integration.py::test_phase_1` (`assert 2 == 1` on the `content_set` count) is a **pre-existing flaky test, unrelated to this `.tekton` change**. It also fails on [`master`](https://github.com/RedHatInsights/vmaas/actions/runs/33513841574) and reproduces intermittently across the `foreman-*` branches. Re-running the job on this same commit passed, confirming it is a race rather than a real regression.

Root cause: the reposcan API tests ([`test/test_reposcan.py`](https://github.com/RedHatInsights/vmaas/blob/master/vmaas/reposcan/test/test_reposcan.py) `test_add_repo_1` / `test_add_repo_1_with_certs`) start an asynchronous `SyncTask` import worker that is only best-effort cancelled. If it is still running when [`test_integration/test_integration.py`](https://github.com/RedHatInsights/vmaas/blob/master/vmaas/reposcan/test_integration/test_integration.py) runs, it can commit the `testing-repo-name` `content_set` row after `test_phase_1`'s `reset_db()`, so the count is 2 instead of 1.

Fix opened in #1942 (against `master`, with a note that it needs backporting to all `foreman-*` branches). This PR's build and hermetic/enterprise-contract checks are green, so it is safe to merge independently.

vkrizan:
@brantleyr can you please rebase?

brantleyr:
@vkrizan @pfreyburg rebased :+1:

vkrizan (approved):

PR #1944, opened 2026-09-08
Stop leftover sync worker between reposcan tests (foreman-5.0)

brantleyr:
Backport of #1942 (merged to `master`) to `foreman-5.0`.

Fixes the intermittent `test_integration.py::test_phase_1` failure (`assert 2 == 1` on the `content_set` count). The reposcan API tests start an asynchronous `SyncTask` import in a spawned worker that is only best-effort cancelled; if it is still running when a later test executes, it can commit a row after that test's `reset_db()`. This adds an autouse teardown that terminates any still-running sync worker at the test boundary so it cannot leak a late commit across tests.

Test-only change (`vmaas/reposcan/conftest.py`).

## Summary by Sourcery

Ensure asynchronous reposcan sync workers are stopped after each test to prevent late database commits from affecting subsequent tests.

Bug Fixes:
- Prevent background sync workers from committing data across reposcan test boundaries, eliminating intermittent database state contamination and related integration test failures.

Tests:
- Add an autouse reposcan test fixture that stops any still-running sync task during teardown.

### DIFF
--- a/vmaas/reposcan/conftest.py
+++ b/vmaas/reposcan/conftest.py
@@ -61,6 +61,28 @@ EXPECTED_CSAF = (
 )
 
 
+@pytest.fixture(autouse=True)
+def _stop_leftover_sync_task():
+    """Ensure no background sync worker survives a test and pollutes the shared DB.
+
+    Reposcan API tests (e.g. test_add_repo_*) POST to /api/v1/repos, which starts an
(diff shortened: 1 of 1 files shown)


brantleyr:
I believe this one is failing because of https://github.com/RedHatInsights/vmaas/pull/1941

jdobes (approved):

PR #1974, opened 2026-09-18
fix(reposcan): make order of repository and CSAF sync interchangeable

jdobes:
RHINENG-31232

## Secure Coding Checklist
- [x] Input Validation
- [x] Output Encoding
- [x] Session Management
- [x] Access Control
- [x] Cryptographic Practices
- [x] Data Protection
- [x] Communication Security
- [x] System Configuration
- [x] Database Security
- [x] File Management
- [x] Memory Management

## Summary by Sourcery

Allow repository and CSAF synchronization to occur in either order without losing package references or metadata.

New Features:
- Allow CSAF package references to be stored before repository package synchronization and link them once repository metadata becomes available.

Bug Fixes:
- Make repository and CSAF synchronization order interchangeable while preserving package associations and enriching previously incomplete package metadata.

Enhancements:
- Introduce a named parsed-NEVRA result for clearer RPM component access.

Tests:
- Add coverage for storing CSAF products before repository catalogs and subsequently enriching their package records.


vmaas/reposcan/database/package_store.py
```diff
@@ -130,6 +149,25 @@ def _populate_packages(self, packages):
...
+                                      source_package_id = coalesce(p.source_package_id, v.source_package_id)
+                                   from (values %s) as v(id, summary, description, source_package_id)
+                                   where p.id = v.id""",
+                               [package[1:] for package in to_update], page_size=len(to_update))
+                self.conn.commit()
+                for package_key, *_ in to_update:
+                    self.package_csaf_no_metadata_map.pop(package_key, None)
+            except Exception:  # pylint: disable=broad-except
+                self.logger.exception("Failure while updating package metadata")
+                self.conn.rollback()
```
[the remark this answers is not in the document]
jdobes: It'll try again on the next run

vmaas/reposcan/database/csaf_store.py
```diff
@@ -122,6 +122,40 @@ def _get_product_attr_id(  # type: ignore[return]
             raise KeyError(f"missing {attr_type}={value}") from err
 
     def _load_product_attr_ids(self, products: model.CsafProducts) -> None:
+        package_names: set[str] = set()
+        fixed_packages: list[dict[str, str | None]] = []
+        parsed_fixed_packages: dict[int, tuple[str, str, str, str, str]] = {}
```
vkrizan: `tuple[str, str, str, str, str]`, why not use a named tuple at this point?

vmaas/reposcan/database/csaf_store.py
```diff
@@ -122,6 +122,40 @@ def _get_product_attr_id(  # type: ignore[return]
...
+            if product.status_id == model.CsafProductStatus.KNOWN_AFFECTED:
+                package_names.add(product.package)
+            elif product.status_id == model.CsafProductStatus.FIXED:
+                try:
+                    parsed_package = rpm.parse_rpm_name(product.package, raise_exception=True)
+                except rpm.RPMParseException as err:
+                    self.logger.debug("Skipping product %s, %s", product, err)
+                    skipped.append(product)
+                    continue
+                parsed_fixed_packages[id(product)] = parsed_package
```
vkrizan: Can you explain the use of Python's internal `id()`? Generally the key would be derived implicitly from the object.

vmaas/reposcan/database/package_store.py
```diff
@@ -19,6 +19,11 @@ def __init__(self):
         self.evr_map = self._prepare_table_map(cols=["epoch", "version", "release"], table="evr")
         self.package_name_map = self._prepare_table_map(cols=["name"], table="package_name")
         self.package_map = self._prepare_table_map(cols=["name_id", "evr_id", "arch_id"], table="package")
+        self.package_csaf_no_metadata_map = self._prepare_table_map(
+            cols=["name_id", "evr_id", "arch_id"],
+            table="package",
+            where="summary IS NULL AND arch_id != (SELECT id FROM arch WHERE name = 'src')",
+        )
```
vkrizan, line 26: (IA): Minor waste, not a bug: package_csaf_no_metadata_map runs a DB query every time a PackageStore is built. This happens even during a CSAF-only run, where the map is never used. Small cost, not worth fixing unless the map grows large.

vmaas/reposcan/database/csaf_store.py
vkrizan, line 1: (IA): Possible risk, worth asking about: before this PR, only repo sync created new package rows. Now CSAF sync can create them too. The INSERT INTO package line (package_store.py line 137) has no ON CONFLICT guard. If CSAF sync and repo sync ever run at the same time and try to add the same new package, one of them could fail with a duplicate-key error. I do not know if these two jobs are already locked to never run together. If they can run at the same time, this is a new small risk. If they are already serialized, this is a non-issue.

vmaas/reposcan/database/package_store.py
jdobes: Should be ok, the PackageStore instance is created only once for each sync type (repo, csaf). It can grow, hopefully not too much. It's a trade-off to quickly detect rows with missing data (added during csaf sync)

vmaas/reposcan/database/csaf_store.py
jdobes: They can't run at the same time. The sync code doesn't support concurrency.

vmaas/reposcan/database/csaf_store.py
jdobes: added support for named tuple

vmaas/reposcan/database/csaf_store.py
jdobes: the `product` is a dataclass and not hashable by default

but the full rpm name is good enough as a key, so I used that instead

vkrizan (approved):
Thanks!

